PR #1, opened 2019-08-21
major code refactoring

FFY00:
Most significant changes are:
  - Each device can now have multiple HID devices
  - Tests now run in a independently from the server
  - Dropped raw_event


src/ratbag_emu/protocol/util/profile.py
```diff
@@ -46,3 +52,9 @@ def dpi(self):
     @dpi.setter
     def dpi(self, dpi):
         self.x_dpi = self.y_dpi = dpi
+
+    def get_dpi(self):
```
bentiss: python has decorators for getters/setters that can be helpful:
```python
class A(object):
  def __init__(self, value):
    self._value = value

  @property
  def value(self):
    return self._value

  @value.setter
  def value(self, new_value):
    assert value > 0
    self._value = value
```

And then you can use it with:

```python
a = A(5)
print(a.value)
a.value = 10
a.value = -6  # <- this should raise an assert
```

src/ratbag_emu/server/device_controller.py
```diff
@@ -18,20 +18,18 @@
 Devices
 '''
 def list_devices():
-    devices = []
+    device_list = []
 
-    for device_id, device in DeviceHandler.devices.items():
+    DeviceHandler.lock.acquire()
+    devices = DeviceHandler.devices.copy()
```
bentiss: not entirely sure if you need to copy it, when you can just release the lock after the for loop.

tests/test_server.py
```diff
@@ -1,46 +1,65 @@
...
+def client():
+    yield Client()
+
+
+def add_device(client, name='steelseries-rival310'):
+    data = {
+        'shortname': name
+    }
+    response = client.post('/devices/add', json=data)
+    return response.json()['id']
```
bentiss: Aren't you supposed to check if the answer is `200`?

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
     data = {
         'shortname': name
     }
-    response = client.post('/devices/add',
-                           data=json.dumps(data),
-                           content_type='application/json')
+    response = client.post('/devices/add', json=data)
     assert response.status_code == 201
-    data = json.loads(response.data)
+    data = response.json()
```
bentiss: `data` is hiding the json `data` description of the device

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
     data = {
         'shortname': name
     }
-    response = client.post('/devices/add',
-                           data=json.dumps(data),
-                           content_type='application/json')
+    response = client.post('/devices/add', json=data)
```
bentiss: before that you should probably ensure that the server has no device named `name`, and then check after the `assert` below that the device is part of the `list_devices` API.

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
     data = {
         'shortname': name
     }
```
bentiss: you should probably in a future PR replace this simple description by the json object we discussed previously that would ensure the creation of the device

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
...
-    response = client.post('/devices/add',
-                           data=json.dumps(data),
-                           content_type='application/json')
+    response = client.post('/devices/add', json=data)
     assert response.status_code == 201
-    data = json.loads(response.data)
+    data = response.json()
     assert 'id' in data
     assert 'shortname' in data
     assert 'name' in data
```
bentiss: you should also ensure that the created device has the correct properties, at the evdev level

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
...
+    assert 'id' in data
+    assert 'shortname' in data
+    assert 'name' in data
+    assert 'input_nodes' in data
 
+    client.delete('/devices/{}'.format(id))
 
-def test_set_dpi(client, dpi_id=0):
+
+def test_dpi(client, dpi_id=0):
```
bentiss: you should have a `test_get_dpi` and a separate `test_set_dpi`, but for that you need to be able to describe entirely your device through json

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
...
     response = client.get('/devices/{}/phys_props/dpi/{}'.format(id, dpi_id))
     assert response.status_code == 200
-    data = json.loads(response.data)
-    assert data == 6666
-
+    assert response.json() == data
 
-def test_get_active_dpi(client):
-    test_get_dpi(client, 'active')
+    client.delete('/devices/{}'.format(id))
```
bentiss: you haven't tested the dpi here. You just tested that the state is somehow consistent.
I need the dpi to be validated by looking at the evdev events. (but maybe this will come later)

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
...
-                          data=json.dumps(data),
-                          content_type='application/json')
+    response = client.put('/devices/{}/phys_props/leds/0'.format(id), json=data)
+    assert response.status_code == 200
+
+    response = client.get('/devices/{}/phys_props/leds/0'.format(id))
     assert response.status_code == 200
+    assert response.json() == data
+
+    client.delete('/devices/{}'.format(id))
```
bentiss: Again, the json description will allow you to properly test the LED, because right now if there is a bug here, we have no idea which function failed.

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
...
+    id = add_device(client, 'generic-hidpp20')
+    input_nodes = []
 
-    DeviceHandler.devices[id].verbose = False
+    start = time()
+    while len(input_nodes) == 0:
+        input_nodes = client.get('/devices/{}'.format(id)).json()['input_nodes']
+        sleep(0.1)
+        if time() > start + 3:
+            break
```
bentiss: yay, polling!
Can't you make the `add_device` call blocking so that you have a full state before requesting the input nodes?

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
...
+        if e.matches(libevdev.EV_REL.REL_X):
+            m.x += e.value
+        elif e.matches(libevdev.EV_REL.REL_Y):
+            m.y += e.value
+
+    dpi = response = client.get('/devices/{}/phys_props/dpi/{}'.format(id, 'active')).json()
+    assert m.x == int(data[0]['action']['x'] * 0.0393700787 * dpi)
+    assert m.y == int(data[0]['action']['y'] * 0.0393700787 * dpi)
+
+    client.delete('/devices/{}'.format(id))
```
bentiss: good, so here you just tested the default dpi.
Now we need to test various other dpis, the buttons, the keys, etc... :)

src/ratbag_emu/protocol/base.py
```diff
@@ -11,24 +11,10 @@
...
-    '''
-    Hardware profile
-
-    Represents the active profile in the hardware.
-    '''
-    hprof = Profile()
-
-    active_profile = None
-    profiles = []
+    hprof = None
```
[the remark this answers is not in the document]
FFY00: > why have you removed the documentation?

What do you mean?

> are you sure this is a good idea to have a class attribute which seems to be instance specific?

I don't think that will generate any issues because it is not shared between devices. I create a profile instance per emulated device. Actually, I want that to be shared between the UHID devices of an emulated device.

src/ratbag_emu/device_handler.py
```diff
@@ -21,6 +21,16 @@ def append_device(device):
         DeviceHandler.devices[DeviceHandler.cur_id] = device
         DeviceHandler.lock.release()
 
+    @staticmethod
+    def wait_for_device_nodes(device_id, timeout=3):
+        start = time()
+        for n, endpoint in DeviceHandler.devices[device_id].endpoints.items():
+            while not endpoint.device_nodes or len(endpoint.device_nodes) == 0:
+                print(f'e.d_n = {endpoint.device_nodes} | len(e.d_n) = {len(endpoint.device_nodes)}')
```
whot: doesn't that always print `[] | 0`?

fwiw, in python any 0 length list evaluates as false, so the length check is superfluous. And, for the future, checking a list for False is more efficient than checking the length for zero since in the former case it doesn't have to actually iterate through the elements.

src/ratbag_emu/protocol/base.py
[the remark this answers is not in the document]
whot: I think there is a rule that "hardware" has to be only ever abbreviated with "hw" anyway. And if there isn't there should totally be one :)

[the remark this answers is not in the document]
whot: > are you sure this is a good idea to have a class attribute which seems to be instance specific?

this is a bit dangerous because python's instance vs class attributes are a bit weird and there's a risk of bus because it won't do what you'd expect:
```
>>> class Foo:
...     bar = None
...     def __init__(self, n):
...             self.bar = n
... 
>>> f = Foo('foo')
>>> b = Foo('bar')
>>> f.bar
'foo'
>>> b.bar
'bar'
>>> Foo.bar
# None
```

IOW, class attributes must be either accessed by class name only or not overwritten in the class with `self`, otherwise they become instance attributes - without changing the class attribute.

This is a python quirk and the general rule I found is "don't declare variables outside `__init__()`". It's very different to other languages (well, Java for example)

Given how you use the `hprof` below, you **definitely** should not declare it as class variable but set it to `None` inside `__init__()`

src/ratbag_emu/protocol/base.py
```diff
@@ -216,15 +202,84 @@ def _send_packets(self, packets, total):
             next_time += 1 / self.hprof.report_rate
         s.run()
 
+
+class BaseDevice():
+    _verbose = True
+
+    endpoints = dict()
```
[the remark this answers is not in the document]
whot: fwiw, more pythonic is `{}` for a dict, `dict()` is afaik only used when you're converting something to a dict (e.g. an iterator)

src/ratbag_emu/protocol/base.py
```diff
@@ -213,18 +199,87 @@ def _send_packets(self, packets, total):
...
+
+    @property
+    def verbose(self):
+        return self._verbose
+
+    @verbose.setter
+    def verbose(self, value):
+        self._verbose = value
+        for n, endpoint in self.endpoints.items():
+            endpoint.verbose = value
```
whot: the dict seems to be overkill anyway (see @bentiss comment) but you can run through the values directly with:
```
for endpoint in self.endpoints.values():
....
```

src/ratbag_emu/protocol/util/profile.py
```diff
@@ -15,13 +15,19 @@ class Profile(object):
     ]
     step = 100
 
-    report_rate = 1000  # Hz
+    active_report_rate = 0
+    report_rates = [ # Hz
+        1000
+    ]
 
+    # TODO: Add proper LED structure
```
whot: for simple structs use [`namedtuple`](https://docs.python.org/3/library/collections.html)

```
Led = namedtuple('Led', ['red', 'green', 'blue'])
l1 = Led(0xff, 0xff, 0xff)
```
Useful if you think you'll eventually implement a class for these anyway, you won't have to change any of the caller code.

src/ratbag_emu/server/device_controller.py
```diff
@@ -18,20 +18,18 @@
...
         if not device:
             continue
 
-        devices.append({
-            'id':           device_id,
-            'name':         device.name,
-            'shortname':    device.shortname,
-            'input_nodes':  device.device_nodes
-        })
+        device_list.append(DeviceHandler.get_device(device_id))
```
whot: you can use list comprehension here

`device_list = [DeviceHandler.get_device(id) for id, device in devices.items() if device is not None]`

though I'm not sure it's that much of an improvement ;)

tests/test_server.py
```diff
@@ -1,46 +1,65 @@
...
 
-    from . import ratbag_emu
 
-    import ratbag_emu.server
-    from ratbag_emu.device_handler import DeviceHandler
+def reload_udev_rules():
+    import subprocess
+
+    subprocess.run('udevadm control --reload-rules'.split())
+    subprocess.run('udevadm hwdb --update'.split())
```
whot: better to call `systemctl hwdb --update`, it actually has a slightly different behaviour. But you don't need this line anyway, you're not updating the hwdb.

tests/test_server.py
```diff
@@ -1,46 +1,65 @@
...
+def udev_rules():
+    import os
+
+    from shutil import copyfile
+    from pathlib import Path
 
-    devices_thread = threading.Thread(target=DeviceHandler.handle)
-    devices_thread.daemon = True
-    devices_thread.start()
+    os.makedirs('/run/udev/rules.d', exist_ok=True)
```
whot: `Path('/run/udev/rules.d').mkdirs(exist_ok=True)` - it's bad form to mix `os.*` and `Path` unless required.

tests/test_server.py
```diff
@@ -1,46 +1,65 @@
...
+    copyfile('rules.d/61-ratbag-emu-ignore-test-devices.rules', '/run/udev/rules.d/61-ratbag-emu-ignore-test-devices.rules')
+    reload_udev_rules()
 
-    yield devices_thread
+    yield
+
+    rules = Path('/run/udev/rules.d/61-ratbag-emu-test-devices.rules')
+    if rules.is_file():
+        rules.unlink()
+        reload_udev_rules()
```
whot: wow, if that works that's a good use of `yield`. Never seen that before. Is `udev_rules()` actually called twice? Which I think is the requirement here.

tests/test_server.py
```diff
@@ -1,46 +1,65 @@
...
+    import os
+
+    from shutil import copyfile
+    from pathlib import Path
 
-    devices_thread = threading.Thread(target=DeviceHandler.handle)
-    devices_thread.daemon = True
-    devices_thread.start()
+    os.makedirs('/run/udev/rules.d', exist_ok=True)
+    copyfile('rules.d/61-ratbag-emu-ignore-test-devices.rules', '/run/udev/rules.d/61-ratbag-emu-ignore-test-devices.rules')
```
whot: you *really* want to have a constant for the final file name. Never `unlink` a file unless it is guaranteed to have the same filename that you used to put it there.

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
...
+    response = client.get('/devices/{}'.format(id))
     assert response.status_code == 200
-    data = json.loads(response.data)
-    assert data == DeviceHandler.devices[id].hprof.dpi[dpi_id if
-        dpi_id != 'active' else DeviceHandler.devices[id].hprof.active_dpi]
+    data = response.json()
+    assert 'id' in data
+    assert 'shortname' in data
+    assert 'name' in data
+    assert 'input_nodes' in data
```
whot: python's unittest has a bunch of helpers like [`TestCase.assertIn`](https://docs.python.org/3/library/unittest.html) which has the added benefit of having useful debug output when it fails.

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
...
+
+    client.delete('/devices/{}'.format(id))
 
 
-def test_device_event_raw_xy(client):
+def test_device_event(client):
     import os
 
-    from time import sleep
+    from time import time, sleep
```
whot: fwiw, localised imports don't really give you that much benefit unless you don't expect that part of the code to be called. Doubly so for full module  imports like `import libevdev` - you'll have to use the module name anyway so just import them at the top of the file.

And in many cases not using `import from` means the namespace provides extra benefit to the reader because it's clearer where a class comes from.

tests/test_server.py
```diff
@@ -52,140 +71,152 @@ def test_add_device(client, name='steelseries-rival310'):
...
+    m = MouseData()
+    for e in events:
+        if e.matches(libevdev.EV_REL.REL_X):
+            m.x += e.value
+        elif e.matches(libevdev.EV_REL.REL_Y):
+            m.y += e.value
+
+    dpi = response = client.get('/devices/{}/phys_props/dpi/{}'.format(id, 'active')).json()
+    assert m.x == int(data[0]['action']['x'] * 0.0393700787 * dpi)
+    assert m.y == int(data[0]['action']['y'] * 0.0393700787 * dpi)
```
whot: for this test it would be more sensible to have constants for `X` and `Y`, use them in `data`  and then re-use them here instead of digging through the dict.

Also, where does the `0.0393...` comes from? General rule is that any not immediately obvious number (i.e. anything other than Pi) needs to have a name or a comment.

tests/test_server.py
```diff
@@ -1,46 +1,65 @@
...
-    import ratbag_emu.server
-    from ratbag_emu.device_handler import DeviceHandler
+def reload_udev_rules():
+    import subprocess
+
+    subprocess.run('udevadm control --reload-rules'.split())
+    subprocess.run('udevadm hwdb --update'.split())
+
+
+@pytest.fixture(scope='session', autouse=True)
```
whot: I *think* now that you're not using the flask wrapper anymore, this is sort of maybe obsolete. And on that note: you'd benefit from putting your tests into a `class TestSomething(unittest.TestCase)` with a `setup` and `teardown` method.

look at the hid-tools tests, they use that approach. You'd have one global setup for the udev rules and one class-specific one that can handle the `add_device` for you.

bentiss:
Not sure why this reply doesn't show up on github, anyway

On Wed, Aug 21, 2019 at 5:54 PM Filipe Laíns <notifications@github.com>
wrote:

> *@FFY00* commented on this pull request.
> ------------------------------
>
> In src/ratbag_emu/protocol/base.py
> <https://github.com/libratbag/ratbag-emu/pull/1#discussion_r316265666>:
>
> > -    '''
> -    Represents an hardware device
> -    '''
> -
> -    protocol = None
> -
> -    '''
> -    Hardware profile
> -
> -    Represents the active profile in the hardware.
> -    '''
> -    hprof = Profile()
> -
> -    active_profile = None
> -    profiles = []
> +    hprof = None
>
> why have you removed the documentation?
>
> What do you mean?
>

in the previous code, `hprof` had a python documentation: `Represents the
active profile in the hardware.`, and now it's gone.


> are you sure this is a good idea to have a class attribute which seems to
> be instance specific?
>
> I don't think that will generate any issues because it is not shared
> between devices. I create a profile instance per emulated device. Actually,
> I want that to be shared between the UHID devices of an emulated device.
>
See @whot's answer, this was not an actual question but a rethorical one to
make you realize that it was not the correct syntax.
If I have to rephrase that question:
"do not use class attribute when you need instance attribute, this is error
prone and forbidden".

Cheers,
Benjamin

src/ratbag_emu/protocol/base.py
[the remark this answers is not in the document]
FFY00: > please use correct names, we are not limited and hprof is not straightforward as hardware_profile. Also, this has the good property of being able to append an s at the end, to tell us if this is a collection or a unique object.

Changed to `hw_profile` as Peter suggested. This variable represents the active settings on the device. The profile list is handled in a different variable (`profiles`). I did this because in some brands (specifically Logitech) the active profile might not be the hardware profile. When you change a profile, the profile settings will be copied to the hardware profile. You can then change a setting directly (ex. using AdjutableDPI). This won't change the active profile, just the hardware profile.

> why have you removed the documentation?

I moved it to `BaseDevice`.

> are you sure this is a good idea to have a class attribute which seems to be instance specific?

I changed that. After I finish dealing with everything I will push the changes for re-review.

src/ratbag_emu/protocol/base.py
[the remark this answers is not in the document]
FFY00: Done.

src/ratbag_emu/protocol/base.py
```diff
@@ -213,18 +199,87 @@ def _send_packets(self, packets, total):
...
+    mouse_endpoint = 0
+    keyboard_endpoint = 0
+    media_endpoint = 0
+
+    def __init__(self, rdesc=None, info=None, name='Generic Device',
+                 shortname='generic'):
+        self.info = info
+        self.name = name
+        self.shortname = shortname
+        self.endpoints[0] = SimulatedDevice(rdesc, info, name, shortname)
```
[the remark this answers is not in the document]
FFY00: TBH, because I don't want to deal with initializing the array, at least right now. I might change that when I implement the json configurator.

src/ratbag_emu/protocol/base.py
```diff
@@ -213,18 +199,87 @@ def _send_packets(self, packets, total):
...
+
+        for n, endpoint in self.endpoints.items():
+            endpoint.hprof = self.hprof
+
+    '''
+    Pass to the correct endpoint
+    '''
+    def create_report(self, data, type=None):
+        self.endpoints[self.mouse_endpoint].create_report(data, type)
```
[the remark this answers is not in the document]
FFY00: This is temporary, it will change when I add the json configurator.

src/ratbag_emu/protocol/util/profile.py
```diff
@@ -15,13 +15,19 @@ class Profile(object):
     ]
     step = 100
 
-    report_rate = 1000  # Hz
+    active_report_rate = 0
+    report_rates = [ # Hz
+        1000
+    ]
```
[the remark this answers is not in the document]
FFY00: Moved to instance attributes.

src/ratbag_emu/protocol/base.py
```diff
@@ -213,18 +199,87 @@ def _send_packets(self, packets, total):
         for i in range(total):
             s.enter(next_time, 1, self.send_raw,
                     kwargs={'data': self.create_report(packets[i], 0x11)})
-            next_time += 1 / self.hprof.report_rate
+            next_time += 1 / self.hprof.get_report_rate()
         s.run()
 
+
+class BaseDevice():
```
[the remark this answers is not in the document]
FFY00: Yup, forgot that.

src/ratbag_emu/protocol/util/profile.py
FFY00: I am aware and I am using that above. Admittedly it's not the best code I've written but `get_dpi()` is just a function to quickly get the active dpi value.

src/ratbag_emu/server/device_controller.py
FFY00: Because `dispatch` will take its time, and it's running on a loop, meaning it will always be running. Copying the device list makes the server much more responsive when we want to modify the device list.

tests/test_server.py
FFY00: `201`, but yes. I forgot.

tests/test_server.py
FFY00: Yes, to avoid constantly writing `response.json()` in the asserts bellow.

tests/test_server.py
FFY00: Why? We can have multiple devices with the same name. That's why we an ID.

src/ratbag_emu/protocol/base.py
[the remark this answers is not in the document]
bentiss: > Changed to hw_profile as Peter suggested. This variable represents the active settings on the device. The profile list is handled in a different variable (profiles). I did this because in some brands (specifically Logitech) the active profile might not be the hardware profile. When you change a profile, the profile settings will be copied to the hardware profile. You can then change a setting directly (ex. using AdjutableDPI). This won't change the active profile, just the hardware profile.

This distinction between hardware profile, active profile and list of profiles seems too tight to a specific vendor.
From what you said, it seems the `hw_profile` is actually the `active_profile`.
This base class is an ideal view of a gaming mouse, and any specifics should be handled in the various protocols.

IMO, it would make more sense to have a list of profiles, and an active profile index, in the same way libratbag exports its profiles.

>>  why have you removed the documentation?
>
> I moved it to BaseDevice.

If it's in a different class, you have to duplicate the documentation. Because you will not inherit it from the other class.

tests/test_server.py
FFY00: Yes but is close enough for us to be able to debug this manually. When we get a better way to test this I will use it.

tests/test_server.py
FFY00: How do I check if I have a full state? I can't reliably predict the amount of device nodes that will show up.

tests/test_server.py
bentiss: it would make sense to have unique names, especially during the tests.
But the whole point of testing that `add a new device` works consists in:
- make sure you do not have the device before hand
- add the device
- then check that the device has been properly added

for the first point, I suggested to make sure you do not have a `name` already, because tests might be running in parallel. Ideally, you should ensure that `list_device` returns an empty list, or even better that no event node exists with that particular unique name.

The ID is nice, but for what we know, there is no guarantees that the server will not just fake a new device, and assign a new id to an existing one.

tests/test_server.py
bentiss: I understand why you need a variable, I am complaining that you are hiding a previous variable. Rename this one to `answer` or something else, please.

src/ratbag_emu/device_handler.py
FFY00: This was leftover debug code. I was running to an odd case where this wasn't true. I can't seem to reproduce it right now so I will just remove it. But I remember clearly, only after I added the length check the code started to behave properly.

tests/test_server.py
bentiss: You *can* know how many devices nodes you need by providing a full json description of your emulated device. You know then that you'll have a mouse, and/or a keyboard. And without having this json description, you already have a static definition of the report descriptors, so you already know before hand which capabilities will need to be present in the event nodes to have a fully ready device.

From the client point of view, you can't really get rid of polling unfortunately, unless your client waits for udev events.  But if the `add_device` is blocking to some extends, then it should be good enough to not hammer the server every 100ms to just get the new devices.

src/ratbag_emu/protocol/base.py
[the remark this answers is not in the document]
FFY00: > This distinction between hardware profile, active profile and list of profiles seems too tight to a specific vendor.
From what you said, it seems the hw_profile is actually the active_profile.
This base class is an ideal view of a gaming mouse, and any specifics should be handled in the various protocols.
IMO, it would make more sense to have a list of profiles, and an active profile index, in the same way libratbag exports its profiles.

We do have that, we just have this extra field.

src/ratbag_emu/protocol/util/profile.py
FFY00: Sure.

tests/test_server.py
FFY00: Done.

tests/test_server.py
FFY00: Yes, the first part is called before staring to run the tests and the last part is called after all tests run.

tests/test_server.py
FFY00: How?

tests/test_server.py
FFY00: Does it really make sense to have a class here? I dislike that approach.

tests/test_server.py
FFY00: > for this test it would be more sensible to have constants for X and Y, use them in data and then re-use them here instead of digging through the dict.

This is just a simple example, what if I add more stuff? I would need to have constants for everything.

> Also, where does the 0.0393... comes from? General rule is that any not immediately obvious number (i.e. anything other than Pi) needs to have a name or a comment.

```
1 mm = 0.0393700787 inch
```

bentiss: > This is just a simple example, what if I add more stuff? I would need to have constants for everything.

yes, you need to add as much constants as possible. But here, it's more that @whot requested you to have the following pseudo-code:
```python
x = 5
y = 6
data = [
        {
            'start': 0,
            'end': 500,
            'action': {
                'type': 'xy',
                'x': x,
                'y': y
            }
        }
    ]
# inject event, read them
assert m.x == int(x * MM_TO_INCHES * dpi)
assert m.y == int(y * MM_TO_INCHES * dpi)
```

FFY00: I still don't agree but sure.

tests/test_server.py
FFY00: I did same as [here](https://gitlab.freedesktop.org/libevdev/python-libevdev/blob/master/examples/evtest.py#L20) and I was able to get the supported events. How can I verify if this is correct?

src/ratbag_emu/protocol/base.py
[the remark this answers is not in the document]
whot: not sure exactly what you mean by initializing the array but maybe this is what you're looking for:  `a = [None] * 20` gives you a 20-sized array with values None. Or you just initialize it to `[]` and `append()` the values, afaict you never need an index that hasn't been initialized yet.

tests/test_server.py
whot: best to add a `# FIXME` then so this bit of information doesn't get lost. Generally good advice, if you're writing code that will require significant changes or has important bits missing, add the fixmes and move on. That way someone else can pick up from it and we all know it's not possible to write code that covers 100% immediately.

tests/test_server.py
whot: /me checks the calendar

Oh, it's 2019. If every 100ms is "hammering" something, you need to upgrade your 286 to something newer, like a 386 ;)

but yeah, I agree, `add_device` should be smarter here.

tests/test_server.py
whot: why? the only noticeable extra work you have is having to move everything right by one indent. benefits are better encapsulation and the ability to re-test things by subclassing. If a test class subclasses another class it will run all tests from the parent, so it's easy to stack tests that way. Every single time I wrote unittests the classes have come in handy.

and the `assertIn` helpers have the advantage of printing the list contents so you can immediately see what else is in that list. `assert` on its own is only useful for "this really mustn't happen" in normal code.

tests/test_server.py
whot: think of it as separation of logic and structure. you want to test that x is 5, so you have a variable for that. Only one place to change it and it all becomes more readable. and sooner or later you likely need to parametrise this anyway - again having a single place to change this is easier.

that's your logic - you want x to be 5 and test for that. The structure is then where you need to use that x, e.g. in this json file and then again in the test. but those shouldn't ever change based on the value of x, so by having a variable you can guarantee that they remain the same structure.

whot: > 1 mm = 0.0393700787 inch

huh, never knew that. I think 25.4 mm per inch is common-enough knowledge, so I've only ever written things as `* 25.4` or `/25.4`. Either way a constant or converter function is safer.

src/ratbag_emu/protocol/base.py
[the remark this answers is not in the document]
FFY00: I want to keep the endpoint numbers intact. I can't assure that they will be loaded in order. The `None` solution seems like a hack and won't work if the device has more than 20 endpoints (not realistic but you never know).

tests/test_server.py
FFY00: > The ID is nice, but for what we know, there is no guarantees that the server will not just fake a new device, and assign a new id to an existing one.

The current implementation does not do this, and I can document this behavior in the API documentation. Would that be enough? I really don't like that approach, I think IDs make more sense, but if you share the same opinion I guess I could change it.

bentiss: > The current implementation does not do this

That's exactly the point of regression tests: you do not trust the underlying code, and you write it as if it was faulty

>  I can document this behavior in the API documentation. Would that be enough?

No, unless you have a test that ensure that the ID is unique, which sums up to exactly what I suggested previously

FFY00:
@whot and @bentiss, I pushed some new code. I am now using a class for tests and I tried to handle the issues I mentioned in the email thread today.

@whot I am not using `unittest` as it messes up pytest. Pytest itself already provides more debug output.

src/ratbag_emu/server/device_controller.py
```diff
@@ -168,13 +166,13 @@ def get_led(device_id, led_id):
        DeviceHandler.devices[device_id] is None:
```
whot: comment on the length check here: general attitude in python is to use exceptions for stuff that shouldn't happen. So more pythonic would be something like this:
```
try:
  device =  DeviceHandler.devices[device_id]
except IndexError:
  device = None
finally:
  if device is None:
    return ...
```

whot: also, I'm a bit confused here (because it's getting late). The `DeviceHandler.devices` is a list but you're using a device ID as index. Which means you need to a) guarantee that the index increments from 0 and b) never remove any device from the list. This seems like it would be better handled by a dictionary.

src/ratbag_emu/protocol/base.py
```diff
@@ -58,7 +44,7 @@ def __init__(self, rdesc=None, info=None, name='Generic Device',
     Prints target message as well as the timestamp
     '''
     def log(self, msg):
-        if self.verbose:
+        if SimulatedDevice.verbose:
```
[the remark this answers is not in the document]
FFY00: I wanted to be able to enable/disable logging per device.

src/ratbag_emu/protocol/hidpp20.py
```diff
@@ -55,8 +55,8 @@ class HIDPP20Device(BaseDevice):
     '''
     def __init__(self, rdesc=None, info=None, feature_table=None,
                  name='Generic HID++ 2.0 Device', shortname='generic-hipp20'):
-        self.protocol = 'HID++ 2.0'
         super().__init__(rdesc, info, name, shortname)
+        self.protocol = 'HID++ 2.0'
```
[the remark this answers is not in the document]
FFY00: I don't think I am using this anywhere right now.

src/ratbag_emu/protocol/util/profile.py
```diff
@@ -35,8 +41,7 @@ def __init__(self, obj=None):
                     self.x_dpi[i] = getattr(res, 'xres')
                     self.y_dpi[i] = getattr(res, 'yres')
 
-                    if hasattr(res, 'is_active'):
-                        if getattr(res, 'is_active'):
+                    if hasattr(res, 'is_active') and getattr(res, 'is_active'):
                             self.active_dpi = i
```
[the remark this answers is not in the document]
FFY00: It is actually

```python

try:
	if res['is_active']:
		self.active_dpi = i
except AttributeError:
	pass
```

which I think is way worse than

```python
if hasattr(res, 'is_active') and getattr(res, 'is_active'):
	self.active_dpi = i
```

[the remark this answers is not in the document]
bentiss: rule of thumb: *never* use `getattr`, `hasattr` and `setattr` unless you are doing introspection, which doesn't seem to be the case here.
So I tend to disagree:
```python
try:
    if res.is_active:
        self.active_dpi = res
except AttributeError:
    pass
```
Is the best pythonic way of sorting this out.

Note that I intentionally used `res` in the assignment, as why keeping the dpi value when you can keep a reference to the object itself.

second rule of thumb: `AttributeError` means a poorly written class, because if you intend to access at some point one attribute, you *must* initialize it to not cripple you code with `try/catch` everywhere.

So, after the second rule, the best way is:
```python
 if res.is_active:
    self.active_dpi = res
```

src/ratbag_emu/server/device_controller.py
FFY00: `DeviceHandler.devices` is a dict.

https://github.com/libratbag/ratbag-emu/blob/07fdaf5b22354798771b20b6263c45be468d824e/src/ratbag_emu/device_handler.py#L11

bentiss: Then you should have a class, and not a dict

bentiss: oops, sorry, I thought you were referring to my comment, ignore that (I haven't gave thought at this part, yet)

src/ratbag_emu/protocol/util/profile.py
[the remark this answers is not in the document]
FFY00: > second rule of thumb: AttributeError means a poorly written class, because if you intend to access at some point one attribute, you must initialize it to not cripple you code with try/catch everywhere.

This value doesn't come from a class, it comes from the json data we receive.

I will use the try block.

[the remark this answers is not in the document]
bentiss: > This value doesn't come from a class, it comes from the json data we receive.

There is something odd here. IIRC, when you parse a json object, you get a plain dict or list.
But `hasattr`, `getattr` and `setattr` should not work on those, especially in the way you are using them.

FWIW, doing `getattr(a, 'constant')` is strictly equivalent to `a.constant`.

So I maintain that you have an object somewhere.

If you really do have a json object, you should use (if you want to convert the `hasattr`:
```python
if 'is_active' in res and res['is_active']:
    self.active_dpi = res
```

but doing the try catch is better, in the python world:
```python
try:
    if res['is_active']:
        self.active_dpi = res
except KeyError:
    pass
```

One last thing: `obj`should probably be renamed `json_object`, and IMO it would be best to convert the json data you get first into a proper python object where you ensure you have your proper API:
```python
class MyObject:
  def __init__(self):
    self.is_active = False
    self.xres = 1000
    self.yres = 1000

  @classmethod
  from_json(cls, json_obj):
    returned_object = cls()
    for key in ['is_active', 'xres', 'yres']:
       if key in in json_obj:
         setattr(returned_object, key) = json_obj[key]

# then convert a json to MyObject:
new_obj = MyObject.from_json(json_data)
```

> I will use the try block.

thanks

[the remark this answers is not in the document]
whot: Doesn't work here but an FYI:  dictionary values can be retrieved with `get()` which defaults to `None` or the given value. So `mydict.get('somekey', 'defaultvalue')` returns either `mydict['somekey']` or `defaultvalue` where the key is not present. That way you don't have to deal with exceptions.

PR #2, opened 2019-08-27
ci: add circleci config

FFY00:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,21 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: archlinux/base
+
+    working_directory: ~/repo
+
+    steps:
+      - run:
+          name: Install dependencies and prepare the system
+          command: |
+            pacman -Syu --noconfirm --noprogressbar hid-tools python-connexion python-pytest python-libevdev git tar
+      - checkout
+      - run:
+          name: Run tests
+          command: |
+            . venv/bin/activate
+            src/main.py &
+            sleep 1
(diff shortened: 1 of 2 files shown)


bentiss:
Hmm, have you tried it actually? `pytest` creates `uhid` devices, so it need root access to `/dev/uhid`.
Unless you are using a VM, this is not safe in a container environment...

FFY00:
You are right, we need to run a VM inside circleci.

FFY00:
Closing this for now.

bentiss:
Dumping some notes here:
- for being able to fully test this, we need a VM
- gitlab.freedesktop.org has a working environment for that: https://gitlab.freedesktop.org/wayland/ci-templates/merge_requests/6
- the gitlab/github integration is only available in premium gitlab, meaning we can't rely on this for the freedesktop instance
- we *could* use a push webhook from github to gitlab to mirror the repo and then use gitlab to send commit statuses through the github API
- this doesn't work much with PR as PR has an event for it, but the data is encapsulated in `HTTP_X_GITHUB_EVENT` and this gets lost if we simply trigger a pipeline from the gitlab API

Solutions seem always to involve running a service somewhere, which is something I do not want to maintain.

So, how about we still use circle CI, and have a special job on top of the containers capable tests that create a new local branch with the MR, add some references in a text file to know which status needs to be updated, pushes it to a gitlab fork of the project, and then we have a `gitlab-ci.yaml` script that creates a qemu, runs the tests and update the status.

Once we have this in place, we can duplicate this for libratbag too.

PR #3, opened 2019-09-03
configure the device based on a json specification

FFY00:


src/ratbag_emu/openapi/ratbag-emu.yaml
```diff
@@ -208,10 +202,6 @@ paths:
           required: true
           schema:
             type: string
-#            anyOf:
-#              - type: integer
-#                format: int32
-#              - type: string
```
bentiss: Can you extract all of these cleanups in a separate commit?

src/ratbag_emu/protocol/base.py
```diff
@@ -29,10 +28,10 @@ def __init__(self, owner, rdesc=None, info=None, name='Generic Device',
 
         self.owner = owner
 
-        self.info = info
+        self.info = owner.info
```
bentiss: might be easier to implement this as a property:
```python
@property
def info(self):
    return self.owner.info
```

src/ratbag_emu/protocol/base.py
```diff
@@ -29,10 +28,10 @@ def __init__(self, owner, rdesc=None, info=None, name='Generic Device',
 
         self.owner = owner
```
bentiss: `self._owner`?

src/ratbag_emu/protocol/base.py
```diff
@@ -29,10 +28,10 @@ def __init__(self, owner, rdesc=None, info=None, name='Generic Device',
...
 
-        self.info = info
+        self.info = owner.info
+        self.shortname = owner.shortname
```
bentiss: could use a property here too

src/ratbag_emu/protocol/base.py
```diff
@@ -127,6 +126,7 @@ def simulate_action(self, actions):
 
             # XY movement
             if action['action']['type'] == 'xy':
+                assert self.hw_profile.get_dpi_value(), 'The device needs a DPI value to emulate mouse movement'
```
bentiss: Having an `assert` in a normal code can be rather dangerous. You should probably rely on an earlier catch of the error

src/ratbag_emu/protocol/base.py
```diff
@@ -201,31 +201,37 @@ def _send_packets(self, packets, total):
 
 
 class BaseDevice(object):
-    def __init__(self, rdesc=None, info=None, name='Generic Device',
-                 shortname='generic'):
-        self.info = info
-        self.name = name
-        self.shortname = shortname
+    def __init__(self, obj):
```
bentiss: I'd keep the existing normal python definition of an object, and add a class method `from_json()` that would create a new object based on a json description

src/ratbag_emu/protocol/base.py
```diff
@@ -201,31 +201,37 @@ def _send_packets(self, packets, total):
...
+        self.info = tuple(obj['bus_info'])
+        self.name = obj['name']
+        self.shortname = obj['shortname']
         self.endpoints = {}
 
-        self.endpoints[0] = SimulatedDevice(self, rdesc, info, name, shortname)
+        for i, rdesc in obj['report_descriptors'].items():
```
bentiss: The report descriptor should not be embedded in the json object but stored in the protocol class or in the `BaseDevice` class (as an attribute you will be overwriting).

src/ratbag_emu/protocol/base.py
```diff
@@ -201,31 +201,37 @@ def _send_packets(self, packets, total):
...
-
-        self.active_profile = None
         self.profiles = []
 
+        self.active_profile = None
+        if 'profiles' in obj:
+            for i, profile in enumerate(obj['profiles']):
+                self.profiles.append(Profile(profile))
+                if profile['is_active']:
+                    self.active_profile = i
```
bentiss: You are using `enumerate` so you can store the index of the active profile. I'd use a simple for/in loop and keep a reference to the active profile instead:
```python
 for profile in obj['profiles']:
    p = Profile(profile)
    self.profiles.append(p)
    if profile['is_active']:
          self.active_profile = p
```

src/ratbag_emu/protocol/base.py
```diff
@@ -201,31 +201,37 @@ def _send_packets(self, packets, total):
...
+
+        if self.profiles:
+            # Represents the active profile in the hardware.
+            self.hw_profile = copy.deepcopy(self.profiles[self.active_profile])
```
bentiss: this would simplify into:
```python
self.hw_profile = copy.deepcopy(self.active_profile)
```

src/ratbag_emu/protocol/hidpp20.py
```diff
@@ -53,11 +53,9 @@ class HIDPP20Device(BaseDevice):
...
     Init routine
     '''
-    def __init__(self, rdesc=None, info=None, feature_table=None,
-                 name='Generic HID++ 2.0 Device', shortname='generic-hipp20'):
-        super().__init__(rdesc, info, name, shortname)
-        self.protocol = 'HID++ 2.0'
-        self.feature_table = feature_table
+    def __init__(self, obj):
+        super().__init__(obj)
+        self.feature_table = obj['hidpp20']['feature_table']
```
bentiss: having a class method `from_json()` as mentioned above would allow you to not modify this file for no reasons.

src/ratbag_emu/protocol/hidpp20.py
```diff
@@ -86,7 +84,7 @@ def __init__(self, rdesc=None, info=None, feature_table=None,
     '''
     Interface functions
     '''
-    def protocol_send(report_type, device, feature, ase, sw_id, args):
+    def protocol_send(self, report_type, device, feature, ase, sw_id, args):
```
bentiss: looks like a separate bug, so separate commit please

src/ratbag_emu/protocol/steelseries.py
```diff
@@ -34,10 +34,10 @@ class SteelseriesDevice(BaseDevice):
...
     Init routine
     '''
-    def __init__(self, rdesc=None, info=None,name='Generic Steelseries Device',
-                 shortname='generic-steelseries', protocol_version=2):
-        self.protocol = f'Steelseries v{protocol_version}'
-        super().__init__(rdesc, info, name, shortname)
+    def __init__(self, obj):
+        super().__init__(obj)
+
+        protocol_version = obj['steelseries']['version']
```
bentiss: Again, there should not be any point in modifying steelseries.
It should be converted in a separate commit, when/if there is a need to modify it.

src/ratbag_emu/protocol/steelseries.py
```diff
@@ -102,7 +101,7 @@ def change_dpi(self, command, data, args):
 
         assert dpi_id == 1 or dpi_id == 2, 'Invalid DPI ID'
 
-        self.hw_profile.dpi[dpi_id - 1] = self.hw_profile.step * (dpi_steps + 1)
+        self.hw_profile.dpi[dpi_id - 1] = self.hw_profile.dpi_step * (dpi_steps + 1)
```
bentiss: Can you convert `step` into `dpi_step` in a separate commit, so we do not pollute this commit with millions of different things?

src/ratbag_emu/protocol/util/profile.py
```diff
@@ -1,47 +1,89 @@
...
 
+
 '''
 Imeplemnts a profile with all the mouse settings
 '''
 class Profile(object):
     Led = namedtuple('Led', ['red', 'green', 'blue'])
 
-    def __init__(self, obj=None):
+    class Button(object):
```
bentiss: is there a specific need to have a nested class?

src/ratbag_emu/protocol/util/profile.py
```diff
@@ -1,47 +1,89 @@
...
+        def __init__(action_type, val=None):
+            self.action_type = action_type
+            self.button = None
+            self.special = None
+            self.macro = None
+
+            setattr(self, self.action_type, val)
```
bentiss: usually, when these kind of things are being written, it's because you are not using the classes correctly:

You should define 3 `ButtonAction` classes, one for each type. Then each class would know how to treat its `value`.

src/ratbag_emu/protocol/util/profile.py
```diff
@@ -1,47 +1,89 @@
...
-        ]
-        self.step = 100
+        self.default_dpi = None
+        self.active_dpi = None
+        self.dpi = []
+        if 'resolutions' in obj:
+            for i, dpi in enumerate(obj['resolutions']):
+                self.dpi.append(dpi['xres'])
+                try:
+                    if dpi['is_active']:
```
bentiss: IIRC @whot left a comment explaining that with json you better use the `get`:
```python
if dpi.get('is_active', False):
    self.active_dpi = i
```
This way you can remove all of the `try/catch KeyError`.

src/ratbag_emu/server/device_controller.py
```diff
@@ -46,52 +46,20 @@ def add_device():
...
-                0x1e22
-            ],
-            'Logitech G Pro', shortname))
-    else:
-        return json.dumps(f"Specification '{shortname}' does not exist"), 404
+    driver = connexion.request.json['driver']
+    spec = connexion.request.json['specification']
+
+    try:
+        if driver == 'generic':
```
bentiss: I'd rather not specify any driver for the `generic` one

tests/test_server.py
```diff
@@ -59,17 +59,125 @@ def udev_rules(self):
...
+                        0x25, 0x7f,  # ....Logical Maximum (127)            44
+                        0x75, 0x08,  # ....Report Size (8)                  46
+                        0x95, 0x02,  # ....Report Count (2)                 48
+                        0x81, 0x06,  # ....Input (Data,Var,Rel)             50
+                        0xc0,        # ...End Collection                    52
+                        0xc0,        # ..End Collection                     53
+                        0xc0,        # .End Collection                      54
+                    ]
+                },
+            }
```
bentiss: This above should all be handled in the server, not in the client.
I do not want from libratbag or curl to have to type in the report descriptor, or the VID/PID.

All of these infromation are known before hand either in the `BaseDevice` class, either in each protocol.

tests/test_server.py
```diff
@@ -59,17 +59,125 @@ def udev_rules(self):
...
+                },
+                'profiles': [
+                    {
+                        'is_active': True,
+                    }
+                ],
+                'steelseries': {
+                    'version': 2
+                }
+            }
```
bentiss: Same here, why would a client need to know which specific report descriptor a Steelseries mouse has?

tests/test_server.py
```diff
@@ -118,43 +226,143 @@ def test_get_device(self, client):
...
+                            'dpi_max': 16000
+                        }
+                    ]
+                }
+            ]
+        })
+
+        new_dpi = 6666
+        response = client.put(f'/devices/{id}/phys_props/dpi/{dpi_id}', json=new_dpi)
+        assert response.status_code == 200
```
bentiss: The 2 tests above assess that the server API is correct.
This is good but we also need to assess that the *emulator* works. So we need the evdev node parsing and whatnot

FFY00:
> We surely do not want any client to use any report descriptor with any mouse. The client should ask for:
> 
>     a generic emulated mouse with a particular profile
>     a default logitech GXYZ or Steelseries Something possibly initialized in a specific state
>     that's it

I agree this would be a nice to have but do we really need it right now? The main job of ratbag-emu is to emulate real devices, and for that we will be using the real report descriptors. The exception would be for unittests here in ratbag-emu, it would definitely be nice to be able to just request a generic device and not having pass any report descriptor but AFAIK that's it. For ratbagd, and other device driver projects, we will be emulating real hardware and we will use the real report descriptors exported by that device.

I agree with everything you said, other than the *surely*. IMHO I don't think it's smart to be wasting time on this right now, since it will not have any impact in the main motivation for this project, which is to create a CI for libratbag.

src/ratbag_emu/protocol/base.py
FFY00: Sure.

src/ratbag_emu/protocol/base.py
FFY00: We don't know if the client will want to do mouse movement before. For ex. if the client is emulating a mouse everything will be fine until he tries to do mouse movement. The code that calls `simulate_action` has a try block to deal with the exception.

Maybe I can just skip that action instead of having an assert, but I think this is more verbose and prevents errors in the client.

bentiss:
> I agree with everything you said, other than the surely. IMHO I don't think it's smart to be wasting time on this right now, since it will not have any impact in the main motivation for this project, which is to create a CI for libratbag.

I insist on the *surely*. :)
I am not asking you to put any effort towards a generic report descriptor generator. But I just want the report descriptor to not be in the client, being the test suite, ratbagd or curl.

All I am asking is that your generic report descriptor in the test suite be an attribute of `BaseDevice`, and let each protocol overwrite it when you instantiate a G903 or a G602.

And yes, this will have a tremendous impact on the project, because you do not want every test in libratbag to carry its report descriptor. This is a known data, and independent of libratbag

src/ratbag_emu/protocol/base.py
bentiss: Other solution is to consider defaults for each value. So if I ask a mouse with:
```json
json = '''
    {
      "profiles": [
        { "is_active": true }
      ]
    }'''
```

Why would you prevent me to compute the dpi from evdev?

src/ratbag_emu/protocol/base.py
FFY00: Done.

src/ratbag_emu/protocol/base.py
FFY00: Done.

src/ratbag_emu/protocol/base.py
FFY00: We might want to describe devices that do not have those properties. For example, I might want to emulate a keyboard.

src/ratbag_emu/protocol/base.py
FFY00: That means more code to maintain. I could do it in the feature but I don't think there's any point doing it right now. It would be nice to be able to create device directly from python without having to create a describing object but that is definitely not a requirement, and so I think I should focus on the missing stuff atm.

src/ratbag_emu/protocol/base.py
bentiss: it really doesn't matter.
The `generic` device can be a set of a mouse and a keyboard (so you can send macros too). So you don't need to worry about keyboards  IMO.

Also, note that I am not asking you to be more restrictive, but less strict. If you have a default device with random default values, you are not preventing a user to input random commands. It's then the user responsibility to have something sane.

src/ratbag_emu/protocol/base.py
FFY00: The report descriptor is part of the device definition, I don't think it should be stored separately. If it helps you could think as an initial value, just like the LEDs or DPI values. Or do you think we should have that separately too? And just like before, the use case where we would be modifying the report descriptor is not really included in the main motivation behind ratbag-emu.

src/ratbag_emu/protocol/base.py
FFY00: The main reason behind the existence of an `self.active_profile` field is not to store the active profile, but an index of the active profile as some protocols will need to export that information. The real active profile, which I call hardware profile is stored in `self.hw_profile`. If you really want, I can change this but just keep in mind I am opposed to it.

src/ratbag_emu/protocol/base.py
FFY00: See comment above.

src/ratbag_emu/protocol/base.py
bentiss: Can we stop arguing for everything?
As you might have realized, when we ask you to do some thing, there is usually a good reason behind it, and we do not ask you to do things to make you lose time. However, the time we all spend arguing whether a request is valid or not is time we effectively lose.

So no, in that case, you will not maintain twice the amount of code, and I believe this is the correct architecture.

Not to mention that you need this factory pattern for the various protocols, as you can make the protocol device a subclass of `BaseDevice`, and the factory allows you to create the correct object from the json description.

So no, I stand my ground, please do it as you are told

src/ratbag_emu/protocol/util/profile.py
FFY00: I can move it out. I just did it to keep in sync with `Led` which was already there.

src/ratbag_emu/protocol/base.py
bentiss: No, the report descriptor should not be included in the json sent over the wire.

> If it helps you could think as an initial value, just like the LEDs or DPI values. 

No, that's 2 rather different concepts. One is written in the ROM of the mouse, the others are values that the user can set.

> Or do you think we should have that separately too?

No, and for the simple reason that you can get exactly what we want in https://github.com/libratbag/libratbag/commit/62ac0772f572ba674ed88373169c0587d5089934

> And just like before, the use case where we would be modifying the report descriptor is not really included in the main motivation behind ratbag-emu.

I mentioned *once* about dynamically generating the report descriptor, and I then said that you could bypass this step for now.
I am not talking about generating a report descriptor, I am talking about where to store the information. And the information should be in the server, not in the client.

src/ratbag_emu/protocol/base.py
bentiss: > The main reason behind the existence of an self.active_profile field is not to store the active profile, but an index of the active profile as some protocols will need to export that information.

I am not asking you to store the entire profile, but a reference to it. And if you *really* need the index, you can either recompute it or store the index in the profile itself (but a property would be better IMO).

> The real active profile, which I call hardware profile is stored in self.hw_profile. If you really want, I can change this but just keep in mind I am somewhat opposed to it.

I honestly never understood the concept of the `self.hw_profile` vs `self.active_profile`. There is a difference that I do not get. And if the difference is because some random mouse has one, then we should have the difference in the protocol, not in the generic base class.

So one more thing to do: remove the `self.hw_profile` and convert it to `self.active_profile`. In a separate commit, please.

src/ratbag_emu/protocol/base.py
FFY00: I am sorry. I was not trying to defy you or anything like that. I am not comfortable implementing things when I do not understand the reasoning behind. What I was trying to do here was to start a discussion, not an argument. You are busy, so I am sorry about that. The goal was to make it to a state where there's no friction between us. I still don't totally understand your POV, and I am uncomfortable with that. Maybe you are perfectly right, but I don't get it. And from my experiences it's generally better to discuss things and try to understand the POVs of one another and then make a decision. With that said, I really am sorry. You are much more busy than I am and you probably have things that need your attention more than me. I will try to limit this behavior.

src/ratbag_emu/protocol/base.py
FFY00: > I am not asking you to store the entire profile, but a reference to it. And if you really need the index, you can either recompute it or store the index in the profile itself (but a property would be better IMO).

```python
self.hw_profile = copy.deepcopy(self.profiles[self.active_profile])
```

This is the only line in the code that will make use of it. My motivation was that we don't really need compute the index each time we need it. But I guess that is just a nitpick. I will change it as you requested.

> I honestly never understood the concept of the self.hw_profile vs self.active_profile. There is a difference that I do not get. And if the difference is because some random mouse has one, then we should have the difference in the protocol, not in the generic base class.
>
> So one more thing to do: remove the self.hw_profile and convert it to self.active_profile. In a separate commit, please.

Several protocols allow you to change the RAM settings without touching the ROM. This means the active settings on the mouse (RAM) might not match the active settings on the profile (ROM). Steelseries does it, Logitech does it and I am pretty sure Razer and Corsair do it too.

Do you need me to elaborate more? Do you still think we should remote it? Let me know.

src/ratbag_emu/protocol/util/profile.py
FFY00: Can I make it an if? 3 separate classes seems like overkill to me.

src/ratbag_emu/protocol/util/profile.py
FFY00: Sure.

tests/test_server.py
FFY00: How? How does `HIDPP20Device` know the VID/PID? Do we have config files for each device?

tests/test_server.py
FFY00: Sure.

FFY00:
> I am not asking you to put any effort towards a generic report descriptor generator. But I just want the report descriptor to not be in the client, being the test suite, ratbagd or curl.

Why? Isn't that what we had previously?

src/ratbag_emu/protocol/base.py
bentiss: thanks. I understand your point of view, but these reviews are eating brain power, especially because I am only reviewing things, not writing. Trust me it would be far quicker for me to write the code than explaining things to you.

Anyway, here, you have the following class hierarchy:
```
BaseDevice <-- HIDPP20Device
            |- SteelseriesDevice
```

So instead of making each class having its `__init__` function parsing the json and have the following pseudo code when instantiating:
```python
def create_device(json_obj):
  protocol = json_obj.get('driver', 'generic')
  if protocol == 'hidpp20':
    return HIDPP20Device(json_obj)
  elif protocol == 'steelseries':
    return SteelseriesDevice(json_obj)
  else:
    return BaseDevice(json_obj)

n = create_device(data)
```

You could use the factory pattern:
```python
class BaseDevice(object):
  generic_rdesc = [0x05, 0x01, ...  # complete generic rdesc
  def __init__(self, info, name, rdesc=BaseDevice.generic_rdesc):
    self._info = info
    self._name = name
    self._rdesc = rdesc

  @classmethod
  def from_json(cls, json_obj):
    returned_object = None
    name =  json_obj.get('name', 'Generic device')
    protocol = json_obj.get('driver', 'generic')
    model = json_obj.get('model', 'generic')
    vid = json_obj.get('vid', 0x0001)
    pid = json_obj.get('pid', 0x0001)
    bus = json_obj.get('bus', 0x0003)
    if protocol == 'hidpp20':
      returned_object = HIDPP20Device(model)
    elif protocol == 'steelseries':
      returned_object = SteelseriesDevice(model)
    else:
      returned_object = BaseDevice((bus, vid, pid), name)
    
    # fill in all the other json properties we might see
    return returned_object

class HIDPP20Models(enum):
    G903 = 1
    G600 = 2
    # TODO: from string is missing, I am sure there is something already available

class HIDPP20Device(BaseDevice):
  standard_rdesc =  [0x05, 0x01, ...  # complete hid++20 standard rdesc
  G600_rdesc = [0x05, 0x01, ...  # complete hid++20 G600 specific rdesc
  
  def __init__(self, model):
    pid = None
    rdesc = HIDPP20Device.standard_rdesc
    if model == HIDPP20Models.G903:
      pid = 0xbeef
    elid model == HIDPP20Models.G600:
      pid = 0xdead
      rdesc = HIDPP20Device.G600_rdesc
    else:
      raise UnknownModelException
    
    super().__init__((3, 0x046d, pid), model.name, rdesc)

# and finally, instantiating an object from json is becoming:
n = BaseDevice.from_json(data)
```

The benefit is that all of your functions but `from_json()` are dealing with standard python, and so you can use whatever representation you want.

Also note how each class can have a specific report descriptor without having to forward it on the wire.

Is that better now?

src/ratbag_emu/protocol/base.py
bentiss: I think you are mixing a few concepts here:
- ROM: Read Only Memory -> memory that is write once in the factory and can never be overwritten
- RAM: Random Access Memory -> volatile memory on a chip
- Flash -> high latency memory, but persistent, used to store the profiles.

(I get that you used ROM for Flash)

Anyway, the problem here is we do not want the emulator to replicate what subtleties a device can have but we want an ideal one.

Here, you are talking about the mice that can be put in the software vs hardware control. Software controlled mice are generally using a RAM profile, as there is no point in storing it on the flash given that the software on the host controls it.

But here, it *really* doesn't matter. It's a firmware implementation that needs to be dealt with in the various protocols.

Our ideal device needs to have several profiles and one of them is the active profile. The latency is null, and we do not care whether we are using software or hardware control. We want to have one active profile we can change, and several other profiles we can swap.

But when you deal with Logitech, Steelseries, or Razer, then yes, you need to implement at most all of the subtleties of the firmware. So with a Logitech mouse in the software mode, a protocol request to change the current dpi would call `set_dpi()` on `BaseDevice`, which would change the currently active profile on the emulator. But when you get a request to read the flash, then you should return whatever you kept in the flash (in `HIDPP20Device`), not the active profile.

src/ratbag_emu/protocol/util/profile.py
bentiss: OK then

src/ratbag_emu/protocol/util/profile.py
bentiss: classes do not cost anything in python.

The `if` would be clearer for the reviewer, but not beneficial as the `setattr` is more efficient.

I still maintain we should use subclassing for that particular case. If you really do not want, please add a `TODO` comment and move on

tests/test_server.py
bentiss: see my code example above - yes - yes (but we can hardcode anything in the code)

tests/test_server.py
bentiss: Please add a `TODO` then

src/ratbag_emu/protocol/base.py
whot: @FFY00: when you just implement something as suggested, best to mark the conversation as resolved as well in the interface.

src/ratbag_emu/protocol/base.py
FFY00: > Is that better now?

It is, I just have some concerns.

```
def create_device(json_obj):
  protocol = json_obj.get('driver', 'generic')
  if protocol == 'hidpp20':
    return HIDPP20Device(json_obj)
  elif protocol == 'steelseries':
    return SteelseriesDevice(json_obj)
  else:
    return BaseDevice(json_obj)

n = create_device(data)
```
I think it would be fine to implement something like this as a `classmethod` or `staticmethod` of `BaseDevice`, so no problems here.

Note that in my code I don't pass the json response from the server to `{HIDPP20,SteelSeries}Device`, I pass the specification (`response = {'driver': ..., 'specification': ...}`). I would like to keep it that way as the `driver` parameter is totally irrelevant after we initialize the right class (after we create a `{HIDPP20,Steelseries}Device` object).

```
  def __init__(self, info, name, rdesc=BaseDevice.generic_rdesc):
    self._info = info
    self._name = name
    self._rdesc = rdesc

  @classmethod
  def from_json(cls, json_obj):
    returned_object = None
    name =  json_obj.get('name', 'Generic device')
    protocol = json_obj.get('driver', 'generic')
    model = json_obj.get('model', 'generic')
    vid = json_obj.get('vid', 0x0001)
    pid = json_obj.get('pid', 0x0001)
    bus = json_obj.get('bus', 0x0003)
    if protocol == 'hidpp20':
      returned_object = HIDPP20Device(model)
    elif protocol == 'steelseries':
      returned_object = SteelseriesDevice(model)
    else:
      returned_object = BaseDevice((bus, vid, pid), name)
    
    # fill in all the other json properties we might see
    return returned_object
```
What feels wrong here is that we have two ways of initializing a device and one of them is incomplete. Actually, this code just won't work as it doesn't initialize the device properties (profiles, etc.).

I did some brain storming, here are some alternatives (they basically do the same thing but are slightly different):

  - Make all arguments in init kwargs, the idea here would be that info, name and rdesc would overwrite what is in the specification we passed, even if it's null. That way we could default the specification to an empty dict and allow the user to create `BaseDevice` in a pythonic way.

POC:
```python
  def __init__(self, info=None, name=None, rdesc=None, obj={'profiles': [{'is_active':True,'xres'=1000}]}):
    if info:
      self._info = info
    else:
      # Get info from obj

    if name:
      self._name = name
    else:
      # Get name from obj

    if None:
      self._rdesc = rdesc
    elif 'report_descriptors' in obj:
      # Get info from obj
    else:
      # Use the default report descriptor
```
Usage:
```python
d1 = BaseDevice([0x01, 0x0001, 0x0001], 'Example Device 1')
d2 = BaseDevice([0x01, 0x0001, 0x0001], 'Example Device 1', obj={
  'profiles': [
    ...
  ]
})
d3 = BaseDevice(obj=obj) # Create a generic device directly
d4 = BaseDevice.create_device(res=res) # Unpack {'driver': ..., 'specification': ...} to a device object (BaseDevice, HIDPP20Device, SteelseriesDevice, etc.)
```

  - Having `__init__` create a specification based on the arguments that it received if no specification was provided.

POC:
```python
  def __init__(self, info=None, name=name, profiles=[{'is_active':True,'xres'=1000}], rdesc=BaseDevice.generic_rdesc, obj={}):
    if not obj and info and name:
      obj = {
        'bus_info': info,
        'name': name,
        'shorname': 'generic-device'
        'report_descriptors': [rdesc],
        'profiles': profiles
      }

    # Create the device based on obj
```
Usage:
```python
d1 = BaseDevice([0x01, 0x0001, 0x0001], 'Example Device 1')
d2 = BaseDevice([0x01, 0x0001, 0x0001], 'Example Device 1', profiles=[
  ...
])
d3 = BaseDevice(obj=obj) # Create a generic device directly
d4 = BaseDevice.create_device(res=res) # Unpack {'driver': ..., 'specification': ...} to a device object (BaseDevice, HIDPP20Device, SteelseriesDevice, etc.)
```

```python
class HIDPP20Device(BaseDevice):
  standard_rdesc =  [0x05, 0x01, ...  # complete hid++20 standard rdesc
  G600_rdesc = [0x05, 0x01, ...  # complete hid++20 G600 specific rdesc
  
  def __init__(self, model):
    pid = None
    rdesc = HIDPP20Device.standard_rdesc
    if model == HIDPP20Models.G903:
      pid = 0xbeef
    elid model == HIDPP20Models.G600:
      pid = 0xdead
      rdesc = HIDPP20Device.G600_rdesc
    else:
      raise UnknownModelException
    
    super().__init__((3, 0x046d, pid), model.name, rdesc)
```
I **rerally** dislike this approach. It will work when we have 2, 5, 10... devices but it is not scalable. Optimally we would be able to emulate all Logitech, Steeelseries, etc. devices and that is not compatible with this approach, at least in a sane way. We would have to add every device manually to the code which is both time consuming and not user friedndly.

I purpose that instead of hardcoding this we have a config system based on the `shortname` field in the specification. This way ratbag-emu would be perfectly scalable. We could then have a script that automatically extracts the necessary information from devices, for ex. in HID++ 2.0 devices it would extract the info, name, report descriptor(s) and feature table. We could save those configs in this repo or in a separate repo.

src/ratbag_emu/protocol/base.py
FFY00: Yes, my bad. I meant Flash, not ROM.

> But here, it really doesn't matter. It's a firmware implementation that needs to be dealt with in the various protocols.

I was trying to ease the implementation of those features in the specific protocol classes as it might be a bit tricky, and we want people to contribute to the project. Having support for this in the backend would be mice IMO but I guess I can change it.

src/ratbag_emu/protocol/steelseries.py
FFY00: But these modifications come from a change in the API, this would introduce a breaking commit. I agree this commit is a mess but I thought this specific change was correct, I can change it though.

src/ratbag_emu/protocol/util/profile.py
FFY00: Okay, that is fine by me. I still don't understand your reasoning. If and when you have a bit of time to explain please do. If I am not wrong these fields don't need any differences in the processing of the data. Actually, I think it would be best to just convert this into a `namedtuple` like `Led`.

src/ratbag_emu/protocol/base.py
bentiss: I think "trying to ease the implementation of those features in the specific protocol classes" is taking the problem backward. We should have a simple and reliable mouse/keyboard/device simulator that does one thing only: emulate events.

We are already handling profiles in this simulator, and I am more and more thinking it should probably be handled in a separate protocol (the generic one). So if the `BaseDevice` class has only the current state in memory and no other state, it would be the responsibility of the protocol to handle profiles, software profiles, hardware profiles and the rest. There might be code duplication in the protocol to handle the profiles, but we can deal with that later.

Furthermore, if I were to write the HID++2.0 hardware profiles feature, I do not want to sync my state with the emulator, and I'd be actually happier to not have to do it. Whenever I receive a get profiles command, I would just return my flash and forget about the rest.

FFY00: > and I am more and more thinking it should probably be handled in a separate protocol

This is perfectly fine by me. We can have ratbagd-like profiles in the a `GenericDevice`. `BaseDevice` would just have a `self.profile` field.

src/ratbag_emu/protocol/base.py
bentiss: > Note that in my code I don't pass the json response from the server to {HIDPP20,SteelSeries}Device, I pass the specification (response = {'driver': ..., 'specification': ...}). I would like to keep it that way as the driver parameter is totally irrelevant after we initialize the right class (after we create a {HIDPP20,Steelseries}Device object).

Well, simply said, I do not want the children class to deal with json, at all.

> What feels wrong here is that we have two ways of initializing a device and one of them is incomplete.

No, you only have one way of initializing a device: the python way. The json needs to be parsed and analyzed in only one function.

> Actually, this code just won't work as it doesn't initialize the device properties (profiles, etc.).

Yes, I haven't written the full code, but gave you an excerpt so you understand my point of view

Regarding your alternatives, I do not like them because you keep the json in the subclasses.

For having written python code in a big project I wish there were type checking for the classes, and I was only using regular positional arguments with (or without) default values.
With json propagated to the subclasses, you can't even rely on `flake8` to ensure your classes are sane, and the immediate benefit you see here is going to be massively overcome next time you need to do a tiny change in the json specification.

> I **really** dislike this approach. It will work when we have 2, 5, 10... devices but it is not scalable.

[I guess it was *really*]

Well, it's a POC, until you get the emulator and the tests running. Once you start implementing HID++ features, you'll need a different way of storing those tables and report descriptors. But honestly for what we are doing, hardcoding the tables in python files is the easiest.

> We would have to add every device manually to the code which is both time consuming and not user friendly.

Yes, this is the exact purpose of what we are doing here: we can not rely on automating things to support new devices, we *need* to ensure a new device is properly emulated.

src/ratbag_emu/protocol/hidpp20.py
```diff
@@ -186,3 +184,165 @@ def IFeatureSet(self, data, ase, args):
...
+            0x1891,
+            0x18a1,
+            0x1801,
+            0x18b1,
+            0x1df3,
+            0x1e00,
+            0x1eb0,
+            0x1863,
+            0x1e22
+        ]
```
bentiss: Not sure if github just stripped the file, but you want to call `super().__init__()` here.

FFY00: I forgot.

src/ratbag_emu/protocol/base.py
```diff
@@ -199,33 +197,38 @@ def _send_packets(self, packets, total):
...
+    device_list = {
+        # Logitech
+        'logitech-g-pro-wireless': LogitechGProWirelessDevice
+    }
 
+    def __init__(self):
+        assert self.name, 'Name is missing'
+        assert self.shortname, 'Shortname is missing'
+        assert self.info, 'Bus info is missing'
+        assert self.rdescs, 'Report descriptor(s) is missing'
```
bentiss: You are preventing anyone to instantiate a `BaseDevice` class here, so you are making it abstract in a way.
But I think we should stick to the previous behaviour, with named parameters, because this will give you the python facilities to know when you forgot to update one parameter in a subclass.

FFY00: Sure. I added a `self.initialized` properly to avoid calling `super().__init__()` with the already initialized attributes.

src/ratbag_emu/protocol/base.py
```diff
@@ -199,32 +197,82 @@ def _send_packets(self, packets, total):
...
+                0xc0,        # ...End Collection                    52
+                0xc0,        # ..End Collection                     53
+                0xc0,        # .End Collection                      54
+            ]
+        ], shortname='custom-generic-device'):
+        if not self.initialized:
+            self.name = name
+            self.shortname = shortname
+            self.info = info
+            self.rdescs = rdescs
```
bentiss: this seems way too complex for what we want.
The initial code was correct: `__init__(self, hw_settings, name, info, rdescs)` and the in subclasses, you just call `super().__init__(my_local_hw_settings, my_local_name, my_local_info, my_local_rdescs)`

FFY00: Okay, I will do that.

PR #7, opened 2019-09-12
Fix test suite

### LINKED ISSUE #6: Test suite is broken
since 3149a2b92d871c48685d4d9253b5a95d7a35fc9f at least

bentiss:
`bool({})` returns False, meaning that an empty hw_settings leads to a server fault.

Fixes #6


src/ratbag_emu/server/device_controller.py
```diff
@@ -53,9 +53,11 @@ def add_device():
         DeviceHandler.append_device(device)
 
     # Generic Device
-    elif hw_settings:
+    elif hw_settings is not None:
         DeviceHandler.append_device(BaseDevice(hw_settings,
                                                id=DeviceHandler.cur_id + 1))
+    else:
+        return json.dumps(f'Missing hw_settings parameter'), 404
```
FFY00: Not needed since `hw_settings` is already specified as required in the openapi specification.

bentiss: It is: sending `{'hw_settings': {}` is translated by flask as `hw_settings = None` here, and then `BaseDevice` goes kaboom.

So yes, `hw_settings` is a requirement, but an empty `hw_settings` is tolerated by OpenAPI and you don't want that.

bentiss: Actually, let me correct that, the problem is `hw_settings = {}` was previously interpreted as false. And we did not created the device.

You must be right, but I'd rather add a comment here saying this should not happen or the if/elif feels like we are missing the default case

FFY00: Yeah, it's fine.

PR #11, opened 2019-09-13
main: move src/main.py to __main__.py and fix pkg data

FFY00:


setup.py
```diff
@@ -46,6 +47,6 @@
     },
     test_requires=['pytest'],
     package_data={
-        'openapi': ['src/ratbag_emu/openapi/ratbag-emu.yaml'],
+        'ratbag_emu': ['openapi/ratbag-emu.yaml'],
```
bentiss: this seems unrelated to the "move src/main.py to __main__.py"

It's best to have one change per commit

FFY00:
Rebased this on #10.

ratbag_emu/__main__.py
```diff
@@ -26,7 +27,7 @@ def main():
 
     # Run server
     server = connexion.FlaskApp(__name__,
-                                specification_dir='ratbag_emu/openapi/',
+                                specification_dir=pkg_resources.resource_filename('ratbag_emu', 'openapi/'),
```
bentiss: Is this change really related to the move from `ratbag-emu.py` to `__main__.py`? (hint: I have no ideas)

setup.py
```diff
@@ -39,7 +39,6 @@
         'ratbag_emu.protocol',
         'ratbag_emu.protocol.devices',
         'ratbag_emu.protocol.util',
-        'ratbag_emu.server',
```
bentiss: Looks like this one should be squashed into the previous commit though

ratbag_emu/__main__.py
FFY00: Yes, we install the specification file as package data and `pkg_resources` allows us to fetch that file without going through hops finding the correct path.

bentiss: OK then

PR #14, opened 2019-09-13
protocol: base: generate a custom string id for every device

FFY00:


FFY00:
I still need to export this in the API.

bentiss:
> I still need to export this in the API

IMO easier would be to leave the caller the choice on which unique name to provide.

So:
- you have an entry point `/funny_rodent_name` (or whatever) that just returns the random name
- then caller creates the device and adds `{'name': 'my_funny_name', 'hw_settings':{...}}`

FFY00:
Eh, just a nitpick but I would prefer to lift the burden of naming the device from the client.

What if we just replace the id exported in the API with our new names? This way we don't have to do much refactoring, which would be easier. We basically remove the numeric ids and replace them with this ones.

bentiss:
> What if we just replace the id exported in the API with our new names?

works for me. As long as it allows your current wip in libratbag to work, I'm all for it

FFY00:
@bentiss can you have a last look before merging?

ratbag_emu/protocol/base.py
```diff
@@ -236,12 +234,20 @@ def __init__(self, hw_settings={}, name='Generic Device',
...
         self.shortname = shortname
         self.info = info
         self.rdescs = rdescs
-        self.id = None
+
+        unique = False
+        while not unique:
+            self.id = self.generate_name()
+            for device in device_list.items():
+                if device[1].id == self.id:
```
bentiss: Not sure where this `[1]` comes from.

Can't you just access the `.id` of the `device`?

Also this seems dangerous: you are relying on the names being unique, but that also means that potentially they can be re-used, and this might mess up our test suite (when running more than one in parallel).

FFY00: > Can't you just access the .id of the device?

We can drop the `.items()` and just do `for id in device_list`. This was leftover code.

> Also this seems dangerous: you are relying on the names being unique, but that also means that potentially they can be re-used, and this might mess up our test suite (when running more than one in parallel).

Well, it depends on how we destroy a device. We can simply remove it from the dictionary or set it to `None`. I would prefer the entry to just be removed from the dict but that means that that name can be reused.

The test suite creates a device, uses it, and when it is done with it, destroys it. It should never try to interface/access the device after that. The only place I can see this being a problem is when trying to access if a device was really destroyed or not in the ratbag-emu test suite. Although *really* unlikely it could happen that another device reincarnates with the same ID. To solve this I propose that we allow the user to choose the ID, like you were initially proposing. If an ID is specified we use it, otherwise we generate one. Sounds good?

bentiss: > The test suite creates a device, uses it, and when it is done with it, destroys it. It should never try to interface/access the device after that. The only place I can see this being a problem is when trying to access if a device was really destroyed or not in the ratbag-emu test suite.

There is also the case where 2 clients request a name, the random returns the same value and given that there is no locking, both are given the same name.

> To solve this I propose that we allow the user to choose the ID, like you were initially proposing. If an ID is specified we use it, otherwise we generate one. Sounds good?

I am mostly nitpicking. I believe it would be easier to get the name from the client: it generates a random sha and request it and can then monitor libratbag for this specific sha.

But OTOH, I don't think you'll enter the race conditions I mentioned above, so we can keep the code as it and change it later.

FFY00: > There is also the case where 2 clients request a name, the random returns the same value and given that there is no locking, both are given the same name.

Well, then we just lock `device_list` and the 2nd client will have to wait until the first client is done, assuring they will have different names.

> I am mostly nitpicking. I believe it would be easier to get the name from the client: it generates a random sha and request it and can then monitor libratbag for this specific sha.

This is fine by me but I didn't want to force the client to give an id if possible.

PR #15, opened 2019-09-13
Add initial circle CI

bentiss:
Just a simple flake8 run that gives us some statistic but fail for syntax error.

Note: I took the skeleton of the flake8 commands from the github actions template. Long term I  think we might want to switch to actions so we can put our gitlab API keys to run the full qemu tests in github itself instead of having to deal with CircleCI.


bentiss:
FWIW, I just enabled circle-ci on libratbag/ratbag-emu

FFY00:
Are you sure? Because I did that.

bentiss:
well, it was not enabled in the libratbag organization, so I did, but we still do not see the badges.

bentiss:
heh, according to the dashboard I beat you by 4 seconds

PR #16, opened 2019-09-17
tests: make pytest start ratbag-emu & better internal api

FFY00:
Signed-off-by: Filipe Laíns <lains@archlinux.org>

PR #18, opened 2019-09-20
Run pytest in qemu from gitlab CI

bentiss:
I initially wanted to wait for #16 to be merged first, but looks like it's taking too much time.

This basically enables true CI for this project: whenever we push something, a pipeline is triggered on https://gitlab.freedesktop.org/bentiss/libratbag-tests which will clone the repo, and run a full pytest in qemu.


tests/test_server.py
```diff
@@ -57,7 +57,23 @@ def udev_rules(self):
             rules.unlink()
             self.reload_udev_rules()
 
-    @pytest.fixture(autouse=True)
+    @pytest.fixture(autouse=True, scope='session')
+    def server(self):
+        p = None
+        logs = f'ratbag-emu-log-{strftime("%Y-%m-%d_%H-%M")}'
+        with open(f'{logs}-stdout.txt', 'w') as stdout, \
+             open(f'{logs}-stderr.txt', 'w') as stderr:
```
FFY00: Is there any way for this files to be recovered from the CI?

FFY00:
> I initially wanted to wait for #16 to be merged first, but looks like it's taking too much time.

Sorry about that, I got tied up in other stuff.

> This basically enables true CI for this project: whenever we push something, a pipeline is triggered on gitlab.freedesktop.org/bentiss/libratbag-tests which will clone the repo, and run a full pytest in qemu.

Are github actions enabled? Can we get it to trigger here?

Also, why didn't the normal CI run? You don't seem to have touched it in the PR.

tests/test_server.py
bentiss: sure, I can add those as artifacts.

On this note, I should rewrite the whole `libratabg-tests` history, but I'll be out for the next 2-3 weeks in conferences, so I wanted to get the PR in first.

FFY00: Do you think we can get this working before that? Both in here and in libratbag?

bentiss:
> Sorry about that, I got tied up in other stuff.

no worries. I wrote a better code, so that's fine :) (kidding)

> Are github actions enabled? Can we get it to trigger here?

yes, they are on the entire libratbag organization.
I am not 100% sure it will work for PR, as github is refactoring the way pipelines and checks are running. But for that being checked, I need to enable this in the master branch first.

> Also, why didn't the normal CI run? You don't seem to have touched it in the PR.

CI was run on the original commit https://github.com/bentiss/ratbag-emu/commit/06be5ac30e253735622391df404813a44c3dde7b

I guess this is where we will have a little bit of baby steps to know when the CI will get run.

tests/test_server.py
bentiss: well, too late, I already pushed the button.

I could do it, but I really don't have time right now. I'll try to have it by the end of the day however.

FFY00: No, I meant do you think we can get the CI working, here and in libratbag, before you leave? If you don't have time I can try to port this to libratbag, I don't think it should be that hard.

bentiss: Ci is already passing for this repo.
The setup has been done so that if the only thing we need is install hid-tools, install requirements.txt, run pytest it should be working for libratbag.

However, there is a chance we will need to manually install ratbag-emu for libratbag, so there might be tweaks to do

PR #19, opened 2019-09-20
tests: add simulate() routine

FFY00:
We will need to run this routine in several tests so I am splitting it
to its own function.


FFY00:
Oh-oh, seems like the CI went away with #18.

FFY00:
@bentiss is it fine to trigger on pull requests?

PR #24, opened 2019-09-30
production: make the server ready for production

FFY00:
Running the server in a production environments seems to solve most of
the lag when running the tests, making them more reliable.

```
$ sudo uwsgi --http :8080 --wsgi-file ratbag_emu/__main__.py
```


ratbag_emu/__main__.py
```diff
@@ -20,23 +20,21 @@
...
+# Run server
+server = connexion.FlaskApp(__name__,
+                            specification_dir=pkg_resources.resource_filename('ratbag_emu', 'openapi/'),
+                            debug=False)
+server.add_api('ratbag-emu.yaml',
+                options={"swagger_ui": True},
+                arguments={'title': 'ratbag-emu'},
+                strict_validation=True,
+                validate_responses=True)
+application = server.app
```
bentiss: That's a lot of changes for an effective one line 2 changes:

change your `main` like the following:
```python
def main(port):
  global application
  # Start handling devices
  ...

  server.add_api('...
  application = server.app
  server.run(port=port)
```

FFY00: `uwsgi` will source the python file. right now `main()` is called inside `if __name__ == '__main__':` which is not desirable. We want to delcare everything in the root script and then actually run the flask app in `if __name__ == '__main__':`.

`if __name__ == '__main__':` will not run when `uwsgi` calls so `main()` won't be run, so we need to run it outside and then use `if __name__ == '__main__':` to start the flask server. There's no point delcaring a main function and running it right after, so I just removed the fucntion.

I think my explanation is not clear enough, so here's some code.

What you are proposing:
```python
...

def main(port):
    global application
    # Start handling devices
    devices_thread = threading.Thread(target=DeviceHandler.handle)
    devices_thread.setDaemon(True)
    devices_thread.start()

    # Run server
    server = connexion.FlaskApp(__name__,
                                specification_dir=pkg_resources.resource_filename('ratbag_emu', 'openapi/'),
                                debug=False)
    server.add_api('ratbag-emu.yaml',
                   options={"swagger_ui": True},
                   arguments={'title': 'ratbag-emu'},
                   strict_validation=True,
                   validate_responses=True)
    server.run(port=port)
    application = server.app


if __name__ == '__main__':
    global application

    ...

    try:
        main(ns.port)
    except KeyboardInterrupt:
        pass
    except Exception:
        traceback.print_exc(file=sys.stdout)
        exit(1)
```

In the `uwsgi` context `if __name__ == '__main__':` won't run so `main()` also won't run, we need to run it outside. But if we run it outside `server.run(port=port)` will run, creating a pure flask server in the background, so we need to move that line to `if __name__ == '__main__':`.

```python
...

def main(port):
    global application
    # Start handling devices
    devices_thread = threading.Thread(target=DeviceHandler.handle)
    devices_thread.setDaemon(True)
    devices_thread.start()

    # Run server
    server = connexion.FlaskApp(__name__,
                                specification_dir=pkg_resources.resource_filename('ratbag_emu', 'openapi/'),
                                debug=False)
    server.add_api('ratbag-emu.yaml',
                   options={"swagger_ui": True},
                   arguments={'title': 'ratbag-emu'},
                   strict_validation=True,
                   validate_responses=True)
    application = server.app

main()


if __name__ == '__main__':
    global application

    ...

    try:
        server.run(port=ns.port)
    except KeyboardInterrupt:
        pass
    except Exception:
        traceback.print_exc(file=sys.stdout)
        exit(1)
```

Now this seems silly, we should supress `main()`.

FFY00: Basically the way that `uwsgi` (and other production servers?) works is that we delcare the server in the root script and then run it inside `if __name__ == '__main__':`. This way it can just source the file without running anything.

bentiss: I still want to not have to install `uswgi`, so I still need the `main()` or any other name for that. I find it silly to have a well scoped function and explode it in the global namespace.

So rename `main()` if you want, but keep it as a single function, so that I can run the server without having to care about `uswgi` but so that your laptop works too.

bentiss: And BTW, your new `main()` in your last example could be named `create_server()` and it returns the server, and you just call

```python
server = create_server()
application = server.app

if __name__ == '__main__':
    # FIXME: argparse handling to be able to add the port
    ns = argparse.Namespace()
    ns.port = 8080
    try:
        server.run(port=ns.port)
    except KeyboardInterrupt:
        pass
    except Exception:
        traceback.print_exc(file=sys.stdout)
        exit(1)
```

FFY00: > I still want to not have to install `uswgi`

Well, this patch changes nothing. You still can do that.

But the other points you made make sense. I will change it.

PR #25, opened 2020-01-17
Major re-write

FFY00:
This is based on my discussion with @bentiss.


FFY00:
Coverage report: https://pkgbuild.com/~ffy00/ratbag-emu/coverage/

ratbag_emu/actuator.py
```diff
@@ -0,0 +1,34 @@
+# SPDX-License-Identifier: MIT
+
+import logging
+import typing
+
+from abc import ABC, abstractmethod
+from typing import List
```
bentiss: nitpick: drop this...

ratbag_emu/actuator.py
```diff
@@ -0,0 +1,34 @@
...
+    Represents the firmware of the device
+
+    Transforms actions based on physical properties
+    '''
+    def __init__(self, owner: 'Device'):
+        self.__logger = logging.getLogger('ratbag-emu.actuator')
+
+        self._owner = owner
+
+        self._keys: List[str] = []
```
bentiss: ... and use `typing.List` directly here.

It's never good to use the `from module import foo` syntax as we end up not knowing where each imported object comes from.

ratbag_emu/device.py
```diff
@@ -0,0 +1,272 @@
...
+
+        We assume there's only one endpoint for each type of action (mouse,
+        keyboard, button, etc.) so we send the action to all endpoints. The
+        endpoint will only act on the action if it supports it.
+
+        :param action:  HID action
+        '''
+        for endpoint in self.endpoints:
+            # FIXME: global data (0x11)?
+            endpoint.send(endpoint.create_report(action, 0x11))
```
bentiss: What is `0x11`?

ratbag_emu/device.py
```diff
@@ -0,0 +1,272 @@
...
+
+        packets: Dict[int, EventData] = {}
+
+        report_count = int(round(ms2s(action['duration']) * self.report_rate))
+
+        if not report_count:
+            report_count = 1
+
+        # XY movement
+        if action['type'] == 'xy':
```
bentiss: we should probably use enums for the action types, so we don't mess up with the strings.

ratbag_emu/device.py
```diff
@@ -0,0 +1,272 @@
...
+                    excess in the buffer for it to be sent in the next
+                    report
+                    '''
+                    if abs(diff) >= 1:
+                        if abs(diff) > 127:
+                            diff = 127 if diff > 0 else -127
+                        setattr(packets[i], attr, diff)
+                        real_pixel_buffer[attr] -= diff
+                i += 1
+
```
bentiss: This code segment is too big to be embedded in an other function. Please extract it in an other function `simulate_action_xy`.

Same for the one below: `simulate_action_button`

ratbag_emu/device.py
```diff
@@ -0,0 +1,272 @@
...
+            real_pixel_buffer holds true number of pixels left to send.
+            When using high report rates (ex. 1000Hz) we usually don't
+            have a full pixel to send, that's why we need two variables. We
+            subtract the step to pixel_buffer at each iteration, when the
+            difference between pixel_buffer and real_pixel_buffer is equal
+            or higher than 1 pixel we then send a HID report to the device
+            with that difference (int! we send the int part of the
+            difference) and update real_pixel_buffer to include this
+            changes.
+            '''
```
bentiss: nitpick: using "pixel" is misleading. We are dealing with a physical mouse, so there is no notion of "pixel". "pixel" is a screen/display notion when here we are only dealing with physical properties.

The term `DPI` stands for `Dots Per Inch`, so we rather use `dots` in place of `pixels`

ratbag_emu/device.py
```diff
@@ -0,0 +1,272 @@
...
+            pixel_buffer = self.transform_action(action['data'])
+
+            for attr in ['x', 'y']:
+                assert pixel_buffer[attr] <= 127 * report_count
```
bentiss: where does the magic `127` comes from?

ratbag_emu/device.py
```diff
@@ -0,0 +1,272 @@
...
+                    The max is 127, if this happens we need to leave the
+                    excess in the buffer for it to be sent in the next
+                    report
+                    '''
+                    if abs(diff) >= 1:
+                        if abs(diff) > 127:
+                            diff = 127 if diff > 0 else -127
+                        setattr(packets[i], attr, diff)
+                        real_pixel_buffer[attr] -= diff
+                i += 1
```
bentiss: The way `i` is used is *very* C-like:

```
            packets: List[EventData] = []
            while real_pixel_buffer['x'] != 0:
                packet = EventData()
                packets.append(packet)

                for attr in ['x', 'y']:
                    pixel_buffer[attr] -= step[attr]
                    diff = int(round(real_pixel_buffer[attr] -
                               pixel_buffer[attr]))
                    # FIXME: Read max size from the report descriptor
                    '''
                    The max is 127, if this happens we need to leave the
                    excess in the buffer for it to be sent in the next
                    report
                    '''
                    if abs(diff) >= 1:
                        if abs(diff) > 127:
                            diff = 127 if diff > 0 else -127
                        setattr(packet, attr, diff)
                        real_pixel_buffer[attr] -= diff
```

bentiss: And there are full of magic `127` everywhere. This should be either a constant, either an attribute of the device so we can have mice that reports more than 127 units at a time.

ratbag_emu/device.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # Button
+        elif action['type'] == 'button':
+            for i in range(report_count):
+                if i not in packets:
+                    packets[i] = EventData()
+
+                setattr(packets[i], 'b{}'.format(action['data']['id']), 1)
```
bentiss: Same here, `packets` would be best used as a list, not a dict

ratbag_emu/device.py
```diff
@@ -0,0 +1,272 @@
...
+
+        def send_packets():
+            nonlocal packets
+            s = sched.scheduler(time.time, time.sleep)
+            next_time = 0
+            for i in range(len(packets)):
```
bentiss: with `packets` as a list: `for packet in packets` -> much simpler :)

ratbag_emu/endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+
+from typing import List
+
+from hidtools.uhid import UHIDDevice
+
+if typing.TYPE_CHECKING:
+    from ratbag_emu.device import Device  # pragma: no cover
+
+
+class Endpoint(UHIDDevice):
```
bentiss: I'd rather have the full `hidtools.uhid.UHIDDevice` instead of the `from ... import ...` above

ratbag_emu/device.py
```diff
@@ -0,0 +1,272 @@
...
+                        0x09, 0x31,  # ....Usage (Y)                        40
+                        0x15, 0x81,  # ....Logical Minimum (-127)           42
+                        0x25, 0x7f,  # ....Logical Maximum (127)            44
+                        0x75, 0x08,  # ....Report Size (8)                  46
+                        0x95, 0x02,  # ....Report Count (2)                 48
+                        0x81, 0x06,  # ....Input (Data,Var,Rel)             50
+                        0xc0,        # ...End Collection                    52
+                        0xc0,        # ..End Collection                     53
+                        0xc0,        # .End Collection                      54
+                     ]]):
```
bentiss: Do we really need to have default values for `info` and `rdescs` here?

I think we better not, so that we are actually forced to add those values in the subclasses or when using this one.

If you really need to have one with default values, just subclass this one as `SimpleMouseDevice`.

ratbag_emu/hw_component.py
```diff
@@ -0,0 +1,24 @@
...
+
+if typing.TYPE_CHECKING:
+    from ratbag_emu.device import Device  # pragma: no cover
+
+
+class HWComponent(object):
+    '''
+    Represents a physical hardware component
+
+    This is the "brain" of the device, it is here where we custom logic is
```
bentiss: typo and we better write this as `This is the "brain" of the device. The custom logic is implemented here`

ratbag_emu/util.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def ms2s(ms):
+    return ms / 1000.0
+
+
+class EventData(object):
+    def __init__(self, x=0, y=0):
+        self.x = x
+        self.y = y
```
bentiss: hmm, not sure the generic `EventData` should have references to `x` and `y`. But we can keep this one for now.

tests/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+import shutil  # noqa: 402
+import subprocess  # noqa: 402
+
+from pathlib import Path  # noqa: 402
+from time import sleep  # noqa: 402
+
+from ratbag_emu.util import EventData  # noqa: 402
+
+
+class TestBase(object):
```
bentiss: shouldn't this class be defined in a separate file?

tests/test_device.py
```diff
@@ -0,0 +1,106 @@
...
+        assert received.y == expected.y
+
+    def test_movement_max(self, device, events):
+        '''
+        Make sure we raise an error when we try to move more than possible
+        '''
+        device.report_rate = 1000
+        device.actuators += [
+            SensorActuator(device, dpi=1000)
+        ]
```
bentiss: I have a feeling the `SensorActuator` should be defined statically when creating the device for the test.
Actually, we might even want to have in the future a logic that parses the provided report descriptor, and creates the corresponding actuators.

ratbag_emu/device.py
```diff
@@ -82,7 +82,7 @@ def __init__(self, name: str = 'Generic Device',
 
         self.report_rate = 100
         self.fw = Firmware(self)
-        self.hw: List[HWComponent] = []
+        self.hw: Dict[str, HWComponent] = {}
```
bentiss: Don  you need to update other parts in the code for that?

ratbag_emu/actuator.py
FFY00: I assume `abc`, right?

ratbag_emu/actuator.py
FFY00: Well, I think that's a great position, however I don't think it's necessary when importing type hint structures from `typing`. Having the full module path makes the code clunky and harder to read, especially when the types start getting more complex.

This is actually the standard way to do it, I believe. I had a long browse at python's source code I was not able to find any instance where they just imported the module.

ratbag_emu/device.py
FFY00: I am not sure but it was in the original code and I didn't remove. I am going to see if it is actually required, I was under the impression that it is.

ratbag_emu/device.py
FFY00:
```
# FIXME: Read max size from the report descriptor
```

It's the max value for axis movement. We should get it from the report descriptor instead of hardcoding an arbitrary `127`.

ratbag_emu/endpoint.py
FFY00: Yes. I agree.

ratbag_emu/device.py
FFY00: We don't. We had them in the old class because you asked me to, so I kept them in the rewrite.

tests/__init__.py
FFY00: I don't believe so, this isn't an actual test. This a base class for the tests, it defines a bunch of fixtures and some utility functions. The tests classes should subclass it and define their tests.

ratbag_emu/device.py
FFY00: No, this was the first commit where I started using it. I found that it was better defined as a dict than as a list, so I changed that and implemented what I needed.

tests/test_device.py
FFY00: What do you mean here? Defined in the device constructor?

Populating the actuators based on the report descriptor would only make sense for generic devices.

ratbag_emu/device.py
FFY00: Is there any easy way to extract this from the UHID device? An error is thrown when trying to set a value above the maximum, so we should know the value at some point.

ratbag_emu/device.py
FFY00: It isn't. Removed.

FFY00:
I will fix the max axis value later.

ratbag_emu/actuator.py
bentiss: > Having the full module path makes the code clunky and harder to read, especially when the types start getting more complex.

Problem is here: I never used the `typing` module previously, so I will always wonder *where* this non standard keyword/class comes from.

> This is actually the standard way to do it, I believe. I had a long browse at python's source code I was not able to find any instance where they just imported the module.

This is not an argument: "everybody uses this, so this is the way it should be used".

Anyway. I am fine keeping the `from foo import blah` in this case, if this is something you truly believe will be clearer in the long run.

ratbag_emu/device.py
bentiss: See https://gitlab.freedesktop.org/libevdev/hid-tools/blob/master/tests/test_multitouch.py#L112-117 on how to access the `logical_max` on a specific hid field. (OK this one is about `features`, but it should be working too with `inputs`)

> Is there any easy way to extract this from the UHID device?

Better use the report descriptor directly instead of waiting for `uhid.ko` to create the device.

.github/workflows/pythonapp.yml
```diff
@@ -23,10 +23,14 @@ jobs:
         git clone https://gitlab.freedesktop.org/libevdev/hid-tools.git
         cd hid-tools
         python setup.py install
+    - name: Static analysis with mympy
```
bentiss: typo: `s/mympy/mypy`

ratbag_emu/endpoint.py
```diff
@@ -0,0 +1,103 @@
...
+
+        self._info = owner.info
+        self.rdesc = rdesc
+        self.number = number
+        self.name = f'ratbag-emu {owner.id} ({owner.name}, ' \
+                    f'{self.vid:04x}:{self.pid:04x}, {self.number})'
+
+        self._output_report = self._receive
+
+        self.create_kernel_device()
```
bentiss: nitpick (for a future use): you might want to split the creation of the object and the actual registering to uhid  which will create the kernel virtual device. This is useful when you want to first get your object, check for some properties on it (like which min/max do we have), and then skip a test.

tests/test_device.py
bentiss: > What do you mean here? Defined in the device constructor?

I think so: `Device` could very well go through the report descriptors, and create the `Actuator` objects it can detect.

This comment was because all of your tests have a lot of duplicated statement for those, and part of it can be automated.

> Populating the actuators based on the report descriptor would only make sense for generic devices.

In a way, yes, but you still can pre-populate  the list of Actuators.

But honestly, while thinking about this more, it might not help as well.

[few seconds and a ahah moment later]

OK, I think I know what should be done for generic devices:
- the tests define the `Actuator`
- then while creating the device, based on the list of `Actuators`, we generate the report descriptor.

This can come in in  a later PR.

bentiss:
Alright, please fix the typo and merge.

ratbag_emu/actuator.py
FFY00: > Problem is here: I never used the typing module previously, so I will always wonder where this non standard keyword/class comes from.

I understand.

> This is not an argument: "everybody uses this, so this is the way it should be used".

That was not really my point. I don't really care if other users use it this way, my point was that the Python developers do.

> Anyway. I am fine keeping the from foo import blah in this case, if this is something you truly believe will be clearer in the long run.

A more complex, but not outlandish, type definition like this

```
Union[List[int], Dict[str, int]]
```

would turn into this


```
typing.Union[typing.List[int], typing.Dict[str, int]]
```

which hurts readability a lot.

ratbag_emu/endpoint.py
FFY00: Noted, thanks.

tests/test_device.py
FFY00: > I think so: Device could very well go through the report descriptors, and create the Actuator objects it can detect.
>
> This comment was because all of your tests have a lot of duplicated statement for those, and part of it can be automated.

But it isn't the same statement. I just add what I need to `device.actuators`, all statements are similar but not quite the same. If I wanted the same statement in all of them I could just put it in the `device` fixture declared above.

```python
    @pytest.fixture()
    def device(self):
        d = Device(name=self.name, info=(3, self.vid, self.pid), rdescs=[[
                        # Generic mouse report descriptor
                        0x05, 0x01,  # .Usage Page (Generic Desktop)        0
                        0x09, 0x02,  # .Usage (Mouse)                       2
                        0xa1, 0x01,  # .Collection (Application)            4
                        0x09, 0x02,  # ..Usage (Mouse)                      6
                        0xa1, 0x02,  # ..Collection (Logical)               8
                        0x09, 0x01,  # ...Usage (Pointer)                   10
                        0xa1, 0x00,  # ...Collection (Physical)             12
                        0x05, 0x09,  # ....Usage Page (Button)              14
                        0x19, 0x01,  # ....Usage Minimum (1)                16
                        0x29, 0x03,  # ....Usage Maximum (3)                18
                        0x15, 0x00,  # ....Logical Minimum (0)              20
                        0x25, 0x01,  # ....Logical Maximum (1)              22
                        0x75, 0x01,  # ....Report Size (1)                  24
                        0x95, 0x03,  # ....Report Count (3)                 26
                        0x81, 0x02,  # ....Input (Data,Var,Abs)             28
                        0x75, 0x05,  # ....Report Size (5)                  30
                        0x95, 0x01,  # ....Report Count (1)                 32
                        0x81, 0x03,  # ....Input (Cnst,Var,Abs)             34
                        0x05, 0x01,  # ....Usage Page (Generic Desktop)     36
                        0x09, 0x30,  # ....Usage (X)                        38
                        0x09, 0x31,  # ....Usage (Y)                        40
                        0x15, 0x81,  # ....Logical Minimum (-127)           42
                        0x25, 0x7f,  # ....Logical Maximum (127)            44
                        0x75, 0x08,  # ....Report Size (8)                  46
                        0x95, 0x02,  # ....Report Count (2)                 48
                        0x81, 0x06,  # ....Input (Data,Var,Rel)             50
                        0xc0,        # ...End Collection                    52
                        0xc0,        # ..End Collection                     53
                        0xc0,        # .End Collection                      54
                     ]])

        yield d

        d.destroy(
```

Although your approach would be cleaner I have some doubts, see my next reply.

Also, this is what it would look like in real code.

```python
class MyCustomDevice(Device):
	def __init__(self, ...):
		...
		self.actuators = [
			CustomActuator1(...)
			CustomActuator2(...)
			CustomActuator3(...)
			CustomActuator4(...)
			CustomActuator5(...)
			CustomActuator6(...)
			CustomActuator7(...)
			CustomActuator8(...)
		]
```

And in a test class where we use the same actuator for all devices.

```python
class TestSomething(TestBase):
	@pytest.fixture()
    def device(self):
        d = Device(name=..., info=(...), rdescs=[...])
		d.actuators = [
			CustomActuator1(...)
			CustomActuator2(...)
			CustomActuator3(...)
			CustomActuator4(...)
		]

        yield d

        d.destroy(
```

Or we would just declare a custom device for it.

```python
class TestSomethingElse(TestBase):
	@pytest.fixture()
    def device(self):
        d = MyCustomDevice(...)

        yield d

        d.destroy(
```

The approach would depend on the problem. Honestly, I don't really see us needing the 1st approach (declare the actuators in the `device` fixture).

> In a way, yes, but you still can pre-populate the list of Actuators.
>
> But honestly, while thinking about this more, it might not help as well.
>
> [few seconds and a ahah moment later]
>
> OK, I think I know what should be done for generic devices:
>
>    - the tests define the Actuator
>    - then while creating the device, based on the list of Actuators, we generate the report descriptor.
>
> This can come in in a later PR.

That does make sense, however I have something to ask: which problem are we trying to solve here? Don't get me wrong, that's not a bad idea, it's just that it's not that simple to implement and I get lost on the motivation behind it. The motivation behind ratbag-emu is for us to be able to emulate real devices and for that this is completely irrelevant.

I think automated actuator discovery would be cool to add later, but at this point I don't really see the need. Do you agree?

PR #30, opened 2020-02-12
fix CI

bentiss:
Add a default logging level to `INFO` and hopefully fix the CI

### DIFF
--- a/ratbag_emu/endpoint.py
+++ b/ratbag_emu/endpoint.py
@@ -2,6 +2,7 @@
 
 import logging
 import struct
+import time
 import typing
 
 import hidtools.uhid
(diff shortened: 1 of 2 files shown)


bentiss:
FYI: https://gitlab.freedesktop.org/bentiss/libratbag-tests/-/jobs/1608264 seems like it does the job

bentiss:
Added one commit to also silence the pyudev warnings.

ratbag_emu/__init__.py
```diff
@@ -5,3 +5,7 @@
 from .endpoint import Endpoint  # noqa: 401
 from .firmware import Firmware  # noqa: 401
 from .hw_component import HWComponent  # noqa: 401
+
+import logging
+__logger = logging.getLogger('ratbag-emu')
+__logger.setLevel(logging.INFO)
```
FFY00: Why is this needed? It can be manually enabled. pytest can also manually enable it via cli options.

bentiss: right. But then can we turn this on in the CI? I mean, pytest will hide logs when everything is fine, so better at least have `--log-level=INFO` by default

FFY00: Yes we can. IIRC pytest will output all the logs at the end, if you want to have them at runtime you could do `--log-file=/dev/stdout`. You can also just save the logs to a normal file.

I am not sure we want to enable this by default in the module. I am fine with enabling them by default in the tests if you want.

bentiss: >  I am fine with enabling them by default in the tests if you want.

OK done: 249a71e46618..7c885bef8d5e  master -> master (in libratbag-tests)

bentiss:
Seems like we solve all the nitpicks, merging

PR #31, opened 2020-02-12
github: pass CODECOV_TOKEN to Gitlab CI

FFY00:


bentiss:
Ouch. This is really prone to leak the secret at some point. Anybody who calls `env` in the `gitlab.yml` will log the value. And this will happen sooner or later. 

I know we control a lot more the CI given that it is on a separate repo, but still...

Is there any reasons we can not store this as a secret in gitlab itself?

FFY00:
We can store it there, sure.

bentiss:
Ok, then we should store it as a file in gitlab secrets, and use it as a file in the codecov script like documented in https://docs.codecov.io/docs/about-the-codecov-bash-uploader#section-file

This way, it will never be stored in the `env` and has much less chances of being leaked.

PR #33, opened 2020-02-12
test

FFY00:
This PR is supposed to trigger the gitlab pipeline, to make sure `DIST_EVENT_NUMBER` is passed correctly.


FFY00:
Hum, it appears the gitlab can't set add/modify workflows on PR.

FFY00:
It looks like this might be happening because the github token doesn't have enough permissions. I managed to update it manually, I think the token must have `repo:status` permission. @bentiss can you check your token and restart the job?

PR #36, opened 2020-02-15
Fix unique ID

FFY00:


FFY00:
Let's try to put a lock and see if it is more reliable.

bentiss:
A lock won't help, you are between different processes

FFY00:
Ah yes, sorry. I looked into it and the only way I think we are able to solve it is by writing a local pytest plugin to inject settings to the various nodes.

An alternative would be, instead of keeping a list, checking if the device exists on the system.

bentiss:
> An alternative would be, instead of keeping a list, checking if the device exists on the system.

or use a wider range of possible unique ids so that the collision risk is minimal:
https://gitlab.freedesktop.org/libevdev/hid-tools/blob/master/hidtools/uhid.py#L168

And it's already there in `uhid.py`. You don't really need the name to be unique, because there is a HID property `uniq` used to report the serial number which is unique enough even across processes.

PR #38, opened 2020-02-20
device: add more ids

FFY00:
To mitigate ID collision I added a color section to the device ID with the X11 color names, giving us 121k possible unique IDs. This is not a proper fix but should be fine for now.

### DIFF
--- a/ratbag_emu/device.py
+++ b/ratbag_emu/device.py
@@ -50,30 +50,60 @@ class Device(object):
         Generates a random name
         '''
         device_names = [
-                'mara', 'capybara', 'porcupine', 'paca',
-                'vole', 'woodrat', 'gerbil', 'shrew',
-                'hutia', 'beaver', 'squirrel', 'chinchilla',
-                'rabbit', 'viscacha', 'hare', 'degu',
-                'gundi', 'acouchy', 'nutria', 'paca',
-                'hamster', 'zokor', 'chipmunk', 'gopher',
-                'marmot', 'groundhog', 'suslik', 'agouti',
-                'blesmol',
+            'mara', 'capybara', 'porcupine', 'paca', 'vole', 'woodrat',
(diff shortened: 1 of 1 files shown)


bentiss:
can we take a step back and explain *why* you need unique names when the `UHIDDevice` already has a unique mechanism in place and can match the created nodes from its own virtual HID device?

FFY00:
How exactly were you thinking of using the `uniq` field?

bentiss:
> How exactly were you thinking of using the uniq field?

I can't remember why you need a unique name. As far as I can see, `UHIDDevice` already filters the user space udev nodes attached to it. So I don't think you need a unique name. Unless I am missing something.

So again: why do you need a unique name?

FFY00:
> I can't remember why you need a unique name. As far as I can see, `UHIDDevice` already filters the user space udev nodes attached to it. So I don't think you need a unique name. Unless I am missing something.

I was not aware of this. We can use that to get the hidraw and event nodes then.

The only problem I see is that by removing the unique name we are possibly making it hard to test external projects. libratbag is fine because it exposes the hidraw nodes in the external API but if a project doesn't do this we make it impossible to identify the device. One thing I wanted to do with ratbag-emu was to write a CI for projects like openrazer and then implement a libratbag driver using the emulator.

bentiss:
> The only problem I see is that by removing the unique name we are possibly making it hard to test external projects. libratbag is fine because it exposes the hidraw nodes in the external API but if a project doesn't do this we make it impossible to identify the device. One thing I wanted to do with ratbag-emu was to write a CI for projects like openrazer and then implement a libratbag driver using the emulator.

I think allowing the external project to use a custom name is better than the automatically generated one. Some kernel drivers and possibly some other projects relie on the name of the device to apply quirks, etc... So it's still good to be able to autogenerate a name, but maybe not by default.

FFY00:
Okay, yes. That makes sense.

PR #39, opened 2020-02-24
device: get nodes from UHIDDevice

FFY00:
@bentiss, the tests are failing because `event_nodes` from `UHIDDevice` is always empty, could you investigate?

### DIFF
--- a/ratbag_emu/device.py
+++ b/ratbag_emu/device.py
@@ -91,6 +91,14 @@ class Device(object):
         return self._info
 
     @property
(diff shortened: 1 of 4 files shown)


bentiss:
> @bentiss, the tests are failing because `event_nodes` from `UHIDDevice` is always empty, could you investigate?

That is because you don't give enough time for the `UHIDDevice` to process the udev events.

I am working on a PR to replace this one and #40.

FFY00:
This should provide the missing bits from #42 that were missing from #40.

Closes #41

PR #40, opened 2020-02-24
device: fix closing event nodes when there is an error when sending events

FFY00:
Right now, if an error is thrown while we are sending the events, the
thread that is listeing is  not be properly handled. This causes the
device to be destroyed with the listening thread still active, resulting
in the error below.

This patch makes sure the listening thread is properly stoped and that
the file descriptors are closed before exiting.

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/anubis/git/ratbag-emu/tests/__init__.py", line 77, in collect_events
    evs += list(device.events())
  File "/usr/lib/python3.8/site-packages/libevdev/device.py", line 523, in events
    ev = self._libevdev.next_event(flags)
  File "/usr/lib/python3.8/site-packages/libevdev/_clib.py", line 872, in next_event
    raise OSError(-rc, os.strerror(-rc))
OSError: [Errno 19] No such device
```

### DIFF
--- a/ratbag_emu/device.py
+++ b/ratbag_emu/device.py
@@ -227,15 +227,10 @@ class Device(object):
         elif action['type'] == ActionType.BUTTON:
             self._simulate_action_xy(action, packets, report_count)
 
-        def send_packets() -> None:
-            nonlocal packets
-            s = sched.scheduler(time.time, time.sleep)
-            next_time = 0.0
-            for packet in packets:
-                s.enter(next_time, 1, self.send_hid_action,
-                        kwargs={'action': packet})
-                next_time += 1 / self.report_rate
-            s.run()
-
-        sim_thread = threading.Thread(target=send_packets)
-        sim_thread.start()
+        s = sched.scheduler(time.time, time.sleep)
+        next_time = 0.0
+        for packet in packets:
+            s.enter(next_time, 1, self.send_hid_action,
+                    kwargs={'action': packet})
+            next_time += 1 / self.report_rate
+        s.run()
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -72,13 +72,18 @@ class TestBase(object):
 
         return devices
 
-    def catch_evdev_events(self, device, evdev_devices, callback, wait=1):
-        evs = []
+    @pytest.fixture()
+    def event_data(self, libevdev_event_nodes):
+        received = EventData()
         def collect_events(stop):  # noqa: 306
-            nonlocal evs
+            nonlocal received
             while not stop.is_set():
-                for evdev_device in evdev_devices:
-                    evs += list(evdev_device.events())
+                for evdev_device in libevdev_event_nodes:
+                    for e in evdev_device.events():
+                        if e.matches(libevdev.EV_REL.REL_X):
+                            received.x += e.value
+                        elif e.matches(libevdev.EV_REL.REL_Y):
+                            received.y += e.value
                 sleep(0.001)
 
         stop_event_thread = threading.Event()
@@ -86,20 +91,10 @@ class TestBase(object):
                                         args=(stop_event_thread,))
         event_thread.start()
 
-        callback(device)
+        yield received
 
-        sleep(wait)
         stop_event_thread.set()
         event_thread.join()
-
-        received = EventData()
-        for e in evs:
-            if e.matches(libevdev.EV_REL.REL_X):
-                received.x += e.value
-            elif e.matches(libevdev.EV_REL.REL_Y):
-                received.y += e.value
-
-        return received
 
     def simulate(self, device, events, action):
         def callback(device):
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: MIT
 
 import random
+import time
 
 import pyudev
 import pytest
@@ -78,7 +79,7 @@ class TestDevice(TestDeviceBase):
                 SensorActuator(dpi=1000)
             ]
 
-    def test_mouse_report(self, device, libevdev_event_nodes):
+    def test_mouse_report(self, device, event_data):
         '''
         Test single HID report
         '''
@@ -86,16 +87,13 @@ class TestDevice(TestDeviceBase):
 
         expected = EventData(x, y)
 
-        def callback(device):
-            nonlocal expected
-            device.send_hid_action(expected)
+        device.send_hid_action(expected)
+        time.sleep(0.1)  # give time for the kernel to proccess all events
 
-        received = self.catch_evdev_events(device, libevdev_event_nodes, callback)
+        assert expected.x <= event_data.x <= expected.x
+        assert expected.y <= event_data.y <= expected.y
 
-        assert expected.x <= received.x <= expected.x
-        assert expected.y <= received.y <= expected.y
-
-    def test_movement(self, device, libevdev_event_nodes):
+    def test_movement(self, device, event_data):
         '''
         Test normal mouse movement
         '''
@@ -114,14 +112,15 @@ class TestDevice(TestDeviceBase):
             }
         }
 
-        received = self.simulate(device, libevdev_event_nodes, action)
+        device.simulate_action(action)
+        time.sleep(0.1)  # give time for the kernel to proccess all events
 
         expected = EventData.from_action(dpi, action)
 
-        assert received.x == expected.x
-        assert received.y == expected.y
+        assert event_data.x == expected.x
+        assert event_data.y == expected.y
 
-    def test_movement_max(self, device, libevdev_event_nodes):
+    def test_movement_max(self, device):
         '''
         Make sure we raise an error when we try to move more than possible
         '''
@@ -140,7 +139,7 @@ class TestDevice(TestDeviceBase):
         }
 
         with pytest.raises(AssertionError):
-            self.simulate(device, libevdev_event_nodes, action)
+            device.simulate_action(action)
 
     def test_duplicated_id(self):
         random.seed(0)


FFY00:
> Ouch, very much ouch.
> 
> There are multiple issues with this PR:
> 
>     * you attempt to fix a `OSError: No such device` by closing the already non existing device... (note that the trace comes from `libevdev`, so the input node)
> 
>     * the trace points at `libevdev`, so the error can't come when "sending events"

The issue was that a error was thrown and the thread was not handled. The device would be destroyed with the thread still trying to read from it.

>     * this is far too complex:
>       
>       * why are you using threads?
>       * why do you have a `callback` callback that takes a `device` parameter that you already provide as an argument -> simply call the `callback` before calling your `catch_events()`

Because we need to receive the events we are sending? How else would we do it? Do the event nodes buffer the data indefinitely?

>       * why do you need to open (and close) the input nodes now? Can't they be opened once when you get the udev event? (this will solve any issue you might have with the previous point just above)

I don't understand. We are not listening to any udev events here. `UHIDDevice` is the only code that listens to them. Are you suggesting that we open the nodes in `Device`? I can see how that is not needed all the time. And how would we handle the hidraw nodes? I don't think it make sense to open those.

>       * `catch_events` is a terrible name. It should be at least `catch_evdev_events` because otherwise it takes a while to fihure out what kind of events it catches (udev, input, HID, cpu?)
>       * `device` vs `devices`: one is HID, the other is a collection of `libevdev.Device)

Sure, that can come in a separate PR. This should have been dealt with before the code was merged.

> Summary: please get rid of the threads here, ensure the input nodes are properly opened when instantiating the `UHDDevice`s. This should sort the main issue here.

FFY00:
I updated the commit message explaining the issue.

bentiss:
> > Ouch, very much ouch.
> > There are multiple issues with this PR:
> > ```
> > * you attempt to fix a `OSError: No such device` by closing the already non existing device... (note that the trace comes from `libevdev`, so the input node)
> > 
> > * the trace points at `libevdev`, so the error can't come when "sending events"
> > ```
> 
> The issue was that a error was thrown and the thread was not handled. The device would be destroyed with the thread still trying to read from it.

Sorry, I can't parse what you said.

But this is exactly why threading in general is a bad thing: understanding the whole races that can happen is *hard*.

> 
> > ```
> > * this is far too complex:
> >   
> >   * why are you using threads?
> >   * why do you have a `callback` callback that takes a `device` parameter that you already provide as an argument -> simply call the `callback` before calling your `catch_events()`
> > ```
> 
> Because we need to receive the events we are sending? How else would we do it? Do the event nodes buffer the data indefinitely?

If you open the device node before injecting events to it, then yes, there is a buffer that will hold the incoming data. Have a look at `hid-tools`, there is no threading and it works just fine:
- have the input event nodes opened once when they are created
- inject a few HID events
- read the matching input evdev events
- inject some other HID events
- rinse wash repeat

However, if you inject too many events at once, you will get a `SYN_DROPPED` event from the kernel and the events will be lost.

> 
> > ```
> >   * why do you need to open (and close) the input nodes now? Can't they be opened once when you get the udev event? (this will solve any issue you might have with the previous point just above)
> > ```
> 
> I don't understand. We are not listening to any udev events here. `UHIDDevice` is the only code that listens to them. Are you suggesting that we open the nodes in `Device`? 
Sigh. `UHIDDevice` listens to the `uhid` events and tells you when there is a new `input`, `hid`, `led`, or anything else device that is a child of this virtual HID device.
So then, you can surely think at something to open the evdev node and store them where you need them to be.
Either let `UHIDDevice` (well `Endpoint`) open them for you, or have `Endpoint` call the `Device` (IIRC) to notify it that there is a new input node available.

> I can see how that is not needed all the time.

well, then make sure you only open them when you need. But it could be interesting to keep the node path around for example.
But still, open the nodes before the test start, and close them at the end (a fixture would be nice for example)

> And how would we handle the hidraw nodes? I don't think it make sense to open those.

-EPARSE (why do you need to open the hidraw nodes?)

> 
> > ```
> >   * `catch_events` is a terrible name. It should be at least `catch_evdev_events` because otherwise it takes a while to fihure out what kind of events it catches (udev, input, HID, cpu?)
> >   * `device` vs `devices`: one is HID, the other is a collection of `libevdev.Device)
> > ```
> 
> Sure, that can come in a separate PR. This should have been dealt with before the code was merged.

Nope. If you want to fix this (and me helping out here), this cleanup needs to come in before this "fix". So either add a commit before cleaning this up, either open a separate PR, get it merged first, and then we can come back to this one.

> 
> > Summary: please get rid of the threads here, ensure the input nodes are properly opened when instantiating the `UHDDevice`s. This should sort the main issue here.

FFY00:
> Sorry, I can't parse what you said.
> 
> But this is exactly why threading in general is a bad thing: understanding the whole races that can happen is _hard_.

I explained it better in the commit message.
 

> If you open the device node before injecting events to it, then yes, there is a buffer that will hold the incoming data. Have a look at `hid-tools`, there is no threading and it works just fine:
> 
>     * have the input event nodes opened once when they are created
> 
>     * inject a few HID events
> 
>     * read the matching input evdev events
> 
>     * inject some other HID events
> 
>     * rinse wash repeat
> 
> 
> However, if you inject too many events at once, you will get a `SYN_DROPPED` event from the kernel and the events will be lost.

This was happening before and it is why we introduced threads. `hid-tools` is a bit different than `ratbat-emu`, in `ratbag-emu` we want to be able to emulate real devices, which spit out a lot of events.

I tried to remove the use of threads and our current test case fails. It tries to move 5x5mm in a linear motion during 0.5sec at a 100Hz report rate and I think this is a perfectly reasonable configuration and not high-spec at all.

With that said, we could always spec down the device but I don't think it is a great solution.

The code to send and receive the events is shared between all tests, so we only have one place to maintain thread code. With this patch it should be working properly.

I understand that maintaining thread code is undesirable but considering the implications that we get for not using threads and the fact we only have 1 thread call to maintain, I think it would be fine, no? What do you think?

> well, then make sure you only open them when you need. But it could be interesting to keep the node path around for example.
> But still, open the nodes before the test start, and close them at the end (a fixture would be nice for example)

That is perfectly fine. We have a fixture for the to get the event nodes, we can modify it to open the file descriptors.

> -EPARSE (why do you need to open the hidraw nodes?)

Right now we are only testing the emulation, so we only need the event nodes. When we start testing protocols we will need the hidraw nodes too. We can do the same as discussed above, open them in a fixture.

> Nope. If you want to fix this (and me helping out here), this cleanup needs to come in before this "fix". So either add a commit before cleaning this up, either open a separate PR, get it merged first, and then we can come back to this one.

That is fine.

FFY00:
So, I did the cleanup as you requested and added fixtures for the event nodes file descriptors and libevdev devices.

The only change requested by you that I did not do is remove the use of threads. Please read my reply above and see if you really want to remove them. The only thread call we have is now properly handled -- all the code between thread creation and exit is enclosed in a `try..finally` loop so the function won't return immediately when an error is thrown, the thread will be closed first.

Let me know what you think.

tests/__init__.py
```diff
@@ -49,23 +49,30 @@ def udev_rules(self):
             self.reload_udev_rules()
 
     @pytest.fixture()
-    def events(self, device, endpoint=0):
-        events = []
+    def event_nodes_fd(self, device, endpoint=0):
```
bentiss: nitpick, rename this to `fd_event_nodes`

tests/__init__.py
```diff
@@ -49,23 +49,30 @@ def udev_rules(self):
...
+        fds = []
         for d in pyudev.Context().list_devices(subsystem='input'):
```
bentiss: We should probably change that statement in a future PR to get the list of event nodes from the endpoints themselves, not having to go through pyudev on all devices.

tests/__init__.py
```diff
@@ -49,23 +49,30 @@ def udev_rules(self):
...
-        evdev_devices = []
-        for node in events:
-            fd = open(node, 'rb')
-            fcntl.fcntl(fd, fcntl.F_SETFL, os.O_NONBLOCK)
-            evdev_devices.append(libevdev.Device(fd))
+        for fd in fds:
+            fd.close()
 
+    @pytest.fixture()
+    def event_nodes_libevdev(self, event_nodes_fd):
```
bentiss: nitpick, this looks much better when named `libevdev_event_nodes`.

tests/__init__.py
```diff
@@ -85,11 +85,13 @@ def collect_events(stop):  # noqa: 306
...
         event_thread.start()
 
-        callback(device)
+        try:
+            callback(device)
 
-        sleep(wait)
-        stop_event_thread.set()
-        event_thread.join()
+            sleep(wait)
```
bentiss: I *think* you should better have:

```python
try:
  callback(device)
except Exception:
  pass
else:
  sleep(wait)
finally:
  stop_event_thread.set()
  event_thread.join()
```

And a couple more questions:
- why is the `sleep(wait)` required? (1 second of wait seems long, you should be there only after having sent the whole batch of events, so just wait for the processing of the last few reports)
- can't we also have the evdev reading Thread as a fixture?

FFY00: Why are we suppressing all exceptions? I think we want to throw all unexpected exceptions, no? In fact, we have a test to make sure we throw an exception if the request data is invalid (ex. try to move 10cm in 0.2s at 800dpi and 100Hz: the total report count is not enough for the movement requested) instead of just failing silently and having random tests fail. Does that make sense?

tests/__init__.py
FFY00: Yes, that should be handled in #39.

tests/__init__.py
FFY00: > why is the sleep(wait) required? (1 second of wait seems long, you should be there only after having sent the whole batch of events, so just wait for the processing of the last few reports)

1sec is the default, it can be removed. The only time we don't use the default duration is for the test that is expected an exception for all other cases we use the duration of the action.

> can't we also have the evdev reading Thread as a fixture?

No because we need to do setup first: add the actuators, set report rate, etc. We also have to define the action. Fixtures are run before the test.

bentiss: > > why is the sleep(wait) required? (1 second of wait seems long, you should be there only after having sent the whole batch of events, so just wait for the processing of the last few reports)
> 
> 1sec is the default, it can be removed. The only time we use the default duration is for the test expects an exception, for all other cases we use the duration of the action.

Hmm, so if you need to do a test that moves a mouse for 5 seconds, you will in addition to the callback wait 5 more seconds?

I am still not quite sure how the callback integrates in this function.

> 
> > can't we also have the evdev reading Thread as a fixture?
> 
> No, because we need to do setup first: add the actuators, set report rate, etc. We also have to define the action. Fixtures are run before the test.

Hmm, OK. but you should probably make this a part of the `Endpoint`: whenever you create the endpoint, you open the evdev nodes, start listening to them in the Thread attached to the endpoint, and then your test can be more reliable and simple.

FFY00: > Hmm, so if you need to do a test that moves a mouse for 5 seconds, you will in addition to the callback wait 5 more seconds?
>
> I am still not quite sure how the callback integrates in this function.

Yes, if the action takes 5 seconds, we need to wait 5 seconds. The action is performed in real-time, we need to listen to the action while it is being performed, which takes 5 seconds.

> Hmm, OK. but you should probably make this a part of the Endpoint: whenever you create the endpoint, you open the evdev nodes, start listening to them in the Thread attached to the endpoint, and then your test can be more reliable and simple.

You still need to explicitly perform the action, so the tests would look the same. The only difference is that instead of starting the read thread in `simulate`, we would do it in the in `Endpoint`.

It is also not a great design decision in my opinion. You will only need to listen to the input events sometimes. If we create 10 devices with 3 endpoints we would be creating 30 listening threads. This not seem like a great approach when you only need to get the input events for 2 or 3 devices. The listening thread would also be present all the time, instead of just while we need them. What do you think?

bentiss: > > Hmm, so if you need to do a test that moves a mouse for 5 seconds, you will in addition to the callback wait 5 more seconds?
> > I am still not quite sure how the callback integrates in this function.
> 
> Yes, if the action takes 5 seconds, we need to wait 5 seconds. The action is performed in real-time, we need to listen to the action while it is being performed, which takes 5 seconds.

Oh. Actually, I think what I missed is that the `callback` is starting a separate thread. So we better just have a synchronous `callback` call instead, given that the evdev events are already fetched in their own thread.


> 
> > Hmm, OK. but you should probably make this a part of the Endpoint: whenever you create the endpoint, you open the evdev nodes, start listening to them in the Thread attached to the endpoint, and then your test can be more reliable and simple.
> 
> You still need to explicitly perform the action, so the tests would look the same. The only difference is that instead of starting the read thread in `simulate`, we would do it in the in `Endpoint`.

Nope, see #42

The tests are now:
```
self.simulate(something)
wait()
self.catch_evdev_events()
```
Which is much more readable than the current `callback` solution where we go back and forth.

> 
> It is also not a great design decision in my opinion. You will only need to listen to the input events sometimes. If we create 10 devices with 3 endpoints we would be creating 30 listening threads. This not seem like a great approach when you only need to get the input events for 2 or 3 devices. The listening thread would also be present all the time, instead of just while we need them. What do you think?

#42 creates only one thread per Device. That is something that can be changed, for sure, but it should work for the time being.
Then, listening to evdev nodes is *cheap*. It seems massively CPU consuming right now because the evdev thread in this PR does active polling on the file descriptors. As soon as you limit the rate, you realize that they are nothing.

And if you really don't want to have the thread all the time ready you can always have something like:
```python
device.start_listening_to_evdev_events()
device.simulate_synchronously()
time.sleep(0.1) # give a little bit of time to process the incoming reports by the kernel and evdev
device.stop_listening_to_evdev_events()
device.catch_evdev_events()
```

And of course, if you are worried about opening too many file descriptors at once, you should delay the creation of the kernel device, which is what I have to do in `hid-tools` when some of the tests should be skipped based on the static definition of the report descriptors.

FFY00: > Nope, see #42

Okay, I understand your implementation proposal and the way it is implemented could be accomplished with a fixture. Like you said:

> whenever you create the endpoint, you open the evdev nodes, start listening to them in the Thread attached to the endpoint, and then your test can be more reliable and simple.

You would have a fixture that would start listening to events when the test starts and stop when the test ends.

This was misunderstanding from my part, I should have understood what you meant the first time you described it. Sorry.

FFY00: I pushed a new commit implementing this. I think it's a cleaner implementation than what we have in #42. Please have a look.

FFY00: @bentiss friendly ping, it has been over a week. Is my approach fine or do you still want to do it in `Device`?

ratbag_emu/device.py
```diff
@@ -227,15 +227,10 @@ def simulate_action(self, action: Dict[str, Any], type: int = None) -> None:
...
-
-        sim_thread = threading.Thread(target=send_packets)
-        sim_thread.start()
+        s = sched.scheduler(time.time, time.sleep)
+        next_time = 0.0
+        for packet in packets:
+            s.enter(next_time, 1, self.send_hid_action,
+                    kwargs={'action': packet})
+            next_time += 1 / self.report_rate
+        s.run()
```
bentiss: I'd like to have this change in a separate commit: it took me a while to figure out that you actually made the `simulate_action` synchronous now.

tests/__init__.py
```diff
@@ -72,34 +72,30 @@ def libevdev_event_nodes(self, fd_event_nodes):
...
             while not stop.is_set():
-                for evdev_device in evdev_devices:
-                    evs += list(evdev_device.events())
+                for evdev_device in libevdev_event_nodes:
+                    for e in evdev_device.events():
+                        if e.matches(libevdev.EV_REL.REL_X):
+                            received.x += e.value
+                        elif e.matches(libevdev.EV_REL.REL_Y):
+                            received.y += e.value
+                sleep(0.001)
```
bentiss: I pushed this `sleep` call on master already

tests/test_device.py
```diff
@@ -78,24 +79,21 @@ def test_duplicated_actuators(self, device):
...
         '''
         x = y = 5
 
         expected = EventData(x, y)
 
-        def callback(device):
-            nonlocal expected
-            device.send_hid_action(expected)
+        device.send_hid_action(expected)
+        time.sleep(0.1)
```
bentiss: You need to add a comment explaining why you need the sleep here, especially given that `send_hid_action` should be synchronous.

tests/test_device.py
```diff
@@ -114,14 +112,15 @@ def test_movement(self, device, libevdev_event_nodes):
             }
         }
 
-        received = self.simulate(device, libevdev_event_nodes, action)
+        device.simulate_action(action)
+        time.sleep(0.1)
```
bentiss: comment here too

bentiss:
An other thing I really dislike in the current code is that too many things are in `tests/__init__.py`. The content of this file should be moved elsewhere. But this can be done in a separate PR.

ratbag_emu/device.py
FFY00: Will do.

tests/__init__.py
FFY00: Okay.

FFY00:
> An other thing I really dislike in the current code is that too many things are in `tests/__init__.py`. The content of this file should be moved elsewhere. But this can be done in a separate PR.

Okay. What do you think of shipping the base test class in the module? We would have to re-implement it in libratbag when we start writing tests.

bentiss:
> > An other thing I really dislike in the current code is that too many things are in `tests/__init__.py`. The content of this file should be moved elsewhere. But this can be done in a separate PR.
> 
> Okay. What do you think of shipping the base test class in the module? We would have to re-implement it in libratbag when we start writing tests.

I am not sure I follow you here. I am actually afraid of understanding it.
The tests class should stick to the `tests` folder. We can then decide what to do with libratbag when we get to it.

FFY00:
> I am not sure I follow you here. I am actually afraid of understanding it.
> The tests class should stick to the `tests` folder. We can then decide what to do with libratbag when we get to it.

Okay, then I'll create a `tests/base.py` file and move the base test class there.

PR #42, opened 2020-02-27
evdev rewrite

bentiss:
This should replace #39 and #40:

First 3 commits are from these PR.
Then we take the code from the tests in `hid-tools` in `Endpoint` to have a nice looking storing of the evdev nodes, and finally we handle the thread in `Device`, because why would we need more than one per device*

\* however, we might revisit that when we need to have more than one evdev node, one per application


bentiss:
closing in favour of #40

PR #27, opened 2020-02-11
WIP: docs: embed mermaid class diagram in the docs

bentiss:
2 objectives:
- try out the CI
- ensure we keep a trace of the class diagram

Still a WIP as we need to add some link on how to see the diagram

PR #48, opened 2020-03-23
Introduce `HIDProperty`

FFY00:
Replaces #45

This deals with the issues from #45 in a nicer way. Instead of a HID Actuator, like suggested, we introduce `HIDProperty`. Each endpoint has a list of properties which is populated by parsing the report descriptor.

`HIDProperty` has a `populate` method which receives an action after it has been updated by the device actuators, and populates a list of HID events with the according HID data. This replaces the `Device._simulate_action_*`.

The idea is to subclass `HIDProperty` for each supported HID data type (eg. `AxisProperty`).

This allows us to drop `ActionType` as everything is now dynamically resolved to HID properties.

What do you think @bentiss? Does this resolve your concerns?

### DIFF
--- a/ratbag_emu/endpoint.py
+++ b/ratbag_emu/endpoint.py
@@ -12,10 +12,10 @@ from typing import Any, List, Dict, Optional
 from ratbag_emu.hid_properties.axis import AxisProperty
 from ratbag_emu.util import EventData
 
-if typing.TYPE_CHECKING:
-    from ratbag_emu.actuator import Actuator  # pragma: no cover # noqa: F401
-    from ratbag_emu.device import Device  # pragma: no cover
-    from ratbag_emu.hid_property import HIDProperty  # pragma: no cover
+if typing.TYPE_CHECKING:  # pragma: no cover
+    from ratbag_emu.actuator import Actuator  # noqa: F401
+    from ratbag_emu.device import Device
+    from ratbag_emu.hid_property import HIDProperty
 
 
 class Endpoint(hidtools.uhid.UHIDDevice):  # type: ignore
--- a/ratbag_emu/endpoint.py
+++ b/ratbag_emu/endpoint.py
@@ -7,10 +7,15 @@ import typing
 
 import hidtools.uhid
 
-from typing import List, Optional
+from typing import Any, List, Dict, Optional
+
+from ratbag_emu.hid_properties.axis import AxisProperty
+from ratbag_emu.util import EventData
 
 if typing.TYPE_CHECKING:
+    from ratbag_emu.actuator import Actuator  # pragma: no cover # noqa: F401
     from ratbag_emu.device import Device  # pragma: no cover
+    from ratbag_emu.hid_property import HIDProperty  # pragma: no cover
 
 
 class Endpoint(hidtools.uhid.UHIDDevice):  # type: ignore
@@ -37,6 +42,9 @@ class Endpoint(hidtools.uhid.UHIDDevice):  # type: ignore
         self.number = number
         self.name = f'ratbag-emu {owner.name} ({self.vid:04x}:{self.pid:04x}, {self.number})'
 
+        self._hid_properties: List[HIDProperty] = []
+        self._update_hid_properties()
+
         self._output_report = self._receive
 
         self.create_kernel_device()
@@ -50,6 +58,13 @@ class Endpoint(hidtools.uhid.UHIDDevice):  # type: ignore
     @property
     def uhid_dev_is_ready(self) -> bool:
         return self.udev_device is not None
+
+    def _update_hid_properties(self) -> None:
+        for features in self.parsed_rdesc.input_reports.values():
+            for feature in features:
+                if feature.usage_name in ['X', 'Y']:
+                    self._hid_properties.append(AxisProperty(self, feature.usage_name.lower(),
+                                                             feature.logical_min, feature.logical_max))
 
     def _receive(self, data: List[int], size: int, rtype: int) -> None:
         '''
@@ -106,3 +121,10 @@ class Endpoint(hidtools.uhid.UHIDDevice):  # type: ignore
             return []
 
         return super().create_report(action, global_data)  # type: ignore
+
+    def populate_hid_data(self, action: Dict[str, Any], packets: List[EventData]) -> None:
+        '''
+        Transforms high-level actions to HID data
+        '''
+        for prop in self._hid_properties:
+            prop.populate(action, packets)
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -77,7 +77,7 @@ API User->>Device: simulate_action(action)
 
 Device->>+Device: hid_action = transform_action(action)
 loop actuators
-	Device-->>Actuator: transform(data)
+    Device-->>Actuator: transform(data)
 end
 Device-->>-Device: return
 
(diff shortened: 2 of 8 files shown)


ratbag_emu/endpoint.py
```diff
@@ -57,6 +57,14 @@ def __init__(self, owner: 'Device', rdesc: List[int], number: int):
     def uhid_dev_is_ready(self) -> bool:
         return self.udev_device is not None
 
+    def _update_hid_properties(self, actuators: List['Actuator']) -> None:
```
bentiss: Is this method ever used?

bentiss (changes requested):
Sorry, I am truly lost here and I can't follow you.

There are barely no commit message explaining the changes so I have to re-read all of the code to get what you are changing.
The PR itself has some explanations, but they need to be put in the commits themselves, not just in the PR message. This PR message will eventually get lost, when the commit descriptions aren't (especially if we switch to gitlab in the future).

It seems to me that the HIDProperty class is the equivalent of what I had in mind for the actuator, so I can't figure out the change in naming.

So instead of answering with words, I'll need from you a class diagram.

Which means, I'm going to block this PR until we get #27 sorted out, with an up to date class diagram. Then, this PR will need to have the matching changes in the diagram, and an explanation on how it is now used.

Sorry for being rude, but I am in the dark for understanding the high level logic of all of this, and I have a feeling you are too.

*Edit: s/I am in the black/I am in the dark/*

ratbag_emu/endpoint.py
FFY00: Yes, every time we change the actuators in `Device`. The idea is that the actuators register properties, and then we scan the report descriptors for those fields.

It seems like there was a mix up, I put the code using that function in the 1st commit. I will fix it.

FFY00:
> There are barely no commit message explaining the changes so I have to re-read all of the code to get what you are changing.
> The PR itself has some explanations, but they need to be put in the commits themselves, not just in the PR message. This PR message will eventually get lost, when the commit descriptions aren't (especially if we switch to gitlab in the future).

Right, sorry. I will try to fix that moving forward.

> It seems to me that the HIDProperty class is the equivalent of what I had in mind for the actuator, so I can't figure out the change in naming.

Well, I was not entirely sure what you had in mind for the HID Actuator. It seemed like this but I didn't know if this was exactly what you had in mind. HIDProperty seemed like a better name to me, but we can change it if you want.

> So instead of answering with words, I'll need from you a class diagram.
> 
> Which means, I'm going to block this PR until we get #27 sorted out, with an up to date class diagram. Then, this PR will need to have the matching changes in the diagram, and an explanation on how it is now used.

I was planning to do that next, but yeah, it probably makes more sense to do it now.

FFY00:
We now have docs and I've added descriptions to the commits. I think the descriptions I added should be enough but please let me know if there is something you can't follow.

FFY00:
@bentiss ping

bentiss:
I still don't get the separation between the `Actuator`, and the `HIDProperty`.

I would really like you to try to see if the code is not cleaner if:
- `HIDProperty` inherits from `Actuator`
- `HIDProperty` references the `EndPoint`
- you might have more than one `Actuator` per type, given the report descriptor
- instead of having `transform()` and `populate()`, you just have `transform()`, which gets reimplemented in `HIDProperty` by taking the old code of `populate()`

FFY00:
Before I make a proper reply I just want to clarify some things.

> I still don't get the separation between the `Actuator`, and the `HIDProperty`.
> 
> I would really like you to try to see if the code is not cleaner if:
> 
>     * `HIDProperty` inherits from `Actuator`

This means each `HIDProperty` will inherit from each `Actuator`, right? If not, could you please clarify what you mean?

>     * `HIDProperty` references the `EndPoint`

Do you mean that `HIDProperty` will use `Endpoint`, and `Endpoint` won't use `HIDProperty`? For eg. `HIDProperty` will use `Endpoint` to send the HID report.

>     * you might have more than one `Actuator` per type, given the report descriptor
> 
>     * instead of having `transform()` and `populate()`, you just have `transform()`, which gets reimplemented in `HIDProperty` by taking the old code of `populate()`

This might seem a bit basic but I just want to make sure we are on the same page.

bentiss:
> Before I make a proper reply I just want to clarify some things.
> 
> > I still don't get the separation between the `Actuator`, and the `HIDProperty`.
> > I would really like you to try to see if the code is not cleaner if:
> > ```
> > * `HIDProperty` inherits from `Actuator`
> > ```
> 
> This means each `HIDProperty` will inherit from each `Actuator`, right? If not, could you please clarify what you mean?

Yes, you can have multiple (HID)Actuators.

> 
> > ```
> > * `HIDProperty` references the `EndPoint`
> > ```
> 
> Do you mean that `HIDProperty` will use `Endpoint`, and `Endpoint` won't use `HIDProperty`? For eg. `HIDProperty` will use `Endpoint` to send the HID report.

I mean that the `HIDProperty` should be aware of its endpoint. So that when you ask for a HID translation, you can ask for "give me the HID translation for report number 2", and the HIDProperty will either fill in the report if it is lined to report number 2 or will silently pass if it's not.

> 
> > ```
> > * you might have more than one `Actuator` per type, given the report descriptor
> > 
> > * instead of having `transform()` and `populate()`, you just have `transform()`, which gets reimplemented in `HIDProperty` by taking the old code of `populate()`
> > ```
> 
> This might seem a bit basic but I just want to make sure we are on the same page.

FFY00:
> > > I still don't get the separation between the `Actuator`, and the `HIDProperty`.
> > > I would really like you to try to see if the code is not cleaner if:
> > > ```
> > > * `HIDProperty` inherits from `Actuator`
> > > ```

Okay, the separation is simple.

`Actuators` transform event data:
  - Axis movement, based on DPI
  - Button mappings
  - etc
So, they essentially provide a frontend in the API. The user says "move 5 mm", we know we want to move 5000 dots, the user says "press button 5", we known we want to send the key A. The idea is for them to just transforms values based on the internal configuration.

`HIDProperty` takes the event and transforms it into a list of objects (HID events) to be sent by hidtools. It essentially converts our event into raw HID data.

The event data will be restructured, the current one is just a prototype. My idea was to use the actuators to entirely replace actions, for eg. when we are working with macros, a "press button 6" event would be replaced with a "move 50 dots on the X axis and press KEY_A down for 1sec" event. 

> > This means each `HIDProperty` will inherit from each `Actuator`, right? If not, could you please clarify what you mean?
> 
> Yes, you can have multiple (HID)Actuators.

If we bind them together we would end up with something like `AxisSensorHIDActuator`, `AxisCustomSensorHIDActuator`, `AxisOtherCustomSensorHIDActuator`, etc. they would all have the same logic, inherited from `AxisHIDActuator`, but we would need to define a (HID)Actuator for each type of actuator.

The HID actuators are also tied to the HID properties in the report descriptor (eg. logical max/min), so I think it makes more sense to have a list of them in each endpoint.

I made a diagram, should make things much easier for you to visualize: https://gitlab.freedesktop.org/snippets/971

I think the depth of this diagram shows the design better than the other one.

> > > ```
> > > * `HIDProperty` references the `EndPoint`
> > > ```
> > 
> > 
> > Do you mean that `HIDProperty` will use `Endpoint`, and `Endpoint` won't use `HIDProperty`? For eg. `HIDProperty` will use `Endpoint` to send the HID report.
> 
> I mean that the `HIDProperty` should be aware of its endpoint. So that when you ask for a HID translation, you can ask for "give me the HID translation for report number 2", and the HIDProperty will either fill in the report if it is lined to report number 2 or will silently pass if it's not.

Okay, the question to ask here is who needs to ask `HIDProperty` for that? I only think that makes sense if we implement what you suggested above, for my suggestion it does not really fit. We can come back to this when we reach a conclusion in the discussion above.

FFY00:
@bentiss friendly ping, if you have time.

bentiss:
OK, I'll try to not be too angry there, but I am.

> The event data will be restructured, the current one is just a prototype. My idea was to use the actuators to entirely replace actions, for eg. when we are working with macros, a "press button 6" event would be replaced with a "move 50 dots on the X axis and press KEY_A down for 1sec" event.

On a PR, we work on the current code, not the future one. Either you have something, you submit it and we discuss, but I can not review something which is just in your head.

That being said, nothing I told you prevent your idea to be working.

> If we bind them together we would end up with something like AxisSensorHIDActuator, AxisCustomSensorHIDActuator, AxisOtherCustomSensorHIDActuator, etc. they would all have the same logic, inherited from AxisHIDActuator, but we would need to define a (HID)Actuator for each type of actuator.

Yes, and?

> The HID actuators are also tied to the HID properties in the report descriptor (eg. logical max/min), so I think it makes more sense to have a list of them in each endpoint.

I don't.

> Okay, the question to ask here is who needs to ask HIDProperty for that? I only think that makes sense if we implement what you suggested above, for my suggestion it does not really fit. We can come back to this when we reach a conclusion in the discussion above.

No, there is no question. I asked you to do one thing. I detailed it:
>     * `HIDProperty` inherits from `Actuator`
>     * `HIDProperty` references the `EndPoint`
>     * you might have more than one `Actuator` per type, given the report descriptor
>     * instead of having `transform()` and `populate()`, you just have `transform()`, which gets reimplemented in `HIDProperty` by taking the old code of `populate()`

And all you are doing is not what I asked. So this is getting really tiring.

Make use of subclassing to separate the concerns (the "action" from the HID conversion), and this should give a better and cleaner code. I am not asking something really complex, I am just asking for a change in how the design is.

If you don't like my idea, I don't care. I want you to implement it nonetheless, to show me what it look like, and then we can decide which one is better. But spending countless time just bikeshedding over a design when *every* single decision needs to be justified is time definitively lost where we could have made a lot of progress.

PR #49, opened 2020-03-25
Class diagram

FFY00:


FFY00:
I tried putting this in sphinx but sphinxcontrib-mermaid is not generating the diagrams properly.

bentiss:
> I tried putting this in sphinx but sphinxcontrib-mermaid is not generating the diagrams properly.

to make gitlab renderer happy, I had to add a bunch of `sth` after the `Device-->>-Device:` when there was no data after

FFY00:
Yes, I figured that and made the change locally when trying to get sphinx to work. I will commit it.

FFY00:
Sphinx still doesn't work with that btw.

Here's a preview: https://pkgbuild.com/~ffy00/documentation/ratbag-emu-sphix-demo/design.html

bentiss:
don't worry too much about that, either we add  a CI job that creates the magic snippet on gitlab, either we'll figure it out later :)

bentiss:
I am pretty sure the arrows are inverted in the class diagram, the "inherits" arrow points at the super class, not the other way around. Not 100% sure for aggregations

FFY00:
My local parse accepts just a space after the `:` in the sequence diagram but it seems gitlab is still not happy with that. Added "return" in those cases.

bentiss:
> I am pretty sure the arrows are inverted in the class diagram, the "inherits" arrow points at the super class, not the other way around. Not 100% sure for aggregations

http://www.agilemodeling.com/artifacts/classDiagram.htm -> the inheritance and aggregations are inverted in your graph

FFY00:
Okay, should be good now.

https://gitlab.freedesktop.org/snippets/842

bentiss:
Sorry, I'm a PITA, but if we want to merge this, it needs to not have the bits of #48, and those bits should be added as part of #48 itself, so we can discuss about it.

FFY00:
Nah, you're not a PITA. I started doing that locally but I figured it was porbably best to hear from you about the first commit before I rebased. Otherwise I would have to sync #48 on every change here.

Is the first commit fine?

bentiss:
yes, I think it is.

Don't bother break the other changes in #48, just add a commit at the end. It won't break anything to not have the diagram in sync for every commit

bentiss (approved):
first commit is good, so you can merge it

FFY00:
> Don't bother break the other changes in #48, just add a commit at the end. It won't break anything to not have the diagram in sync for every commit

It's easier for me to just rebase on top of this branch.

